PR #3, opened 2023-05-14
Clean up a couple of things

40thieves:
Removes an unnecessary log and gitignores a Hugo build lockfile. Noticed while experimenting with the codebase.


SallyMcGrath (approved):
ty

PR #6, opened 2023-05-16
Time default should be set in minutes

SallyMcGrath:
The default time was set as an hour eg 0.3 for 20 minutes
It should have been set as minutes eg 20
I have corrected this in the module script
And updated the example blocks generated from this script
None of these blocks will survive content population but it's a bad example, so I have rectified


SallyMcGrath:
I'm just merging as nobody else has context yet really

PR #10, opened 2023-05-21
Add support for CodeMirror editing via a render hook

40thieves:
## Description

Adds support for [CodeMirror 6](https://codemirror.net/) via a [render hook](https://gohugo.io/templates/render-hooks/). So adding something like:

    ```cm6
    console.log('some JS code here')
    ```

will get converted into a CM6 editor:

Screenshot 2023-05-21 at 18 47 41

Opening as a draft as I'd like some feedback on:

1. I've set it up so that you must specifically use `cm6` in the code fence, but I could easily switch this to take over any JS code block
    - What is more tricky is supporting arbitrary languages - partly because CM6 only supports a [certain set of languages out-of-the-box](https://codemirror.net/#languages), and partly because each of these languages would have to be installed separately. However, I will note that [CM6 has explicit support for loading languages on the fly](https://codemirror.net/docs/ref/#language.LanguageDescription.load), so it could be done
2. The general experience of using the CM6 editor. CM6 is extremely customisable, to the point where the default config does almost nothing over plain `<div contenteditable="true">`. I've loaded [the `basicSetup` extensions](https://codemirror.net/docs/ref/#codemirror.basicSetup), which adds most of the features you'd expect, but it's fairly easy to configure how this works.

I'd be more than happy to walk people through CM6 as the docs aren't the most friendly to without some context.

Finally, I should note that I've _not_ implemented a "run this code" button for now.

@SallyMcGrath 


40thieves:
⚠️⚠️⚠️ REVERT BEFORE MERGING! ⚠️⚠️⚠️

I added a temporary commit to demonstrate it in the deploy preview https://github.com/CodeYourFuture/curriculum/pull/10/commits/6e8d297005e3cb36d6248dc2037bef2704484ed2 If you go to the Fundamentals prep page you'll see it show up

40thieves:
I've added another commit (https://github.com/CodeYourFuture/curriculum/pull/10/commits/be3e67bb0850b0932e7ff5b2389dbfa2ff5a614d) which enables the indent via Tab feature. This is [disabled by default due to a11y concerns](https://codemirror.net/examples/tab/) but I think the opt-in behaviour is still relatively accessible (press `Esc` then `Tab` to move focus out of the editor)

SallyMcGrath:
🎉 YAY THANK YOU

I confirm, esc to exit is fine. We might want a hidden heading before the div  in the partial to produce a code block outline (and warning) 

``` <h4 class="is-invisible">Interactive code block</h4>```

So there are some decisions to be made, I think you're saying.

My instinct is to set up the js codemirror with all the stuff we want and have it intercede in all codefences marked js. Then if people specifically don't want code mirror they can leave off the js flag, but all js written in github readmes will be highlighted on github and codemirrored on the website without friction.

If you agree (and you may not) questions to answer:

- what is the performance of a long page with many snippets? 
- what options do we want?

- [ ] - run code
- [ ] - line wrap
- [ ] - line numbers
- [ ] - go to line?
- [ ] - theme one dark, monokai, or vulcan?
- [ ] - autocomplete?

Later, we could make a second bundle for SQL, then Bash, then CSS, etc? And only load those ones on the pages that call them, I would think.

**Comment by 40thieves** [comment]
> We might want a hidden heading before the div in the partial to produce a code block outline (and warning)

Yeah I thought about that and wondered if it'd just be annoying? Especially to have it read out the whole "Esc then Tab" instructions every time.

> if people specifically don't want code mirror they can leave off the js flag, but all js written in github readmes will be highlighted on github and codemirrored on the website without friction

I see where you're coming from, but in this scenario you wouldn't get JS highlighting on the website then. I'm happy to, if we think it's worth the tradeoff.

> what is the performance of a long page with many snippets?

Good point, I'll test that. A single CM6 editor is reasonably fast in my experience, see [this example](https://codemirror.net/examples/million/), although we did have some issues in my day job - possible/probable this is due to some of our customisations though.

> there are some decisions to be made

Yes 🙂

> what options do we want?

One thing I noticed: the default font size is fairly small, so I think could do with bumping up. 

> theme one dark, monokai, or vulcan?

Are you happy with the current theme? It's literally just the [one-dark theme](https://github.com/codemirror/theme-one-dark) that comes bundled in the basicSetup package.

Fwiw, [here's the docs for creating your own theme](https://codemirror.net/docs/ref/#view.EditorView%5Etheme) - it uses the author's [own CSS-in-JS package](https://github.com/marijnh/style-mod)

> go to line?

This is actually [implemented out-of-the-box](https://codemirror.net/docs/ref/#search.searchKeymap) - it's triggered via `Alt-G`. But it's fairly trivial to change it to another binding if wanted. I will say that this was one of the problems we had to solve in my day job - there's not really an obvious consensus around what the binding for go to line should be.

> autocomplete?

There should be _some_ autocomplete working:

Screenshot 2023-05-22 at 17 31 16

I'm honestly not too sure how useful it is in reality. It'd be a bit of a larger job, but we could look at tweaking it.

> Later, we could make a second bundle for SQL, then Bash, then CSS, etc? 

Makes sense - especially if we decide to CM6-ise all JS code blocks.

**Comment by SallyMcGrath** [comment]
> wondered if it'd just be annoying? Especially to have it read out the whole "Esc then Tab" instructions every time.

Literally just "interactive code", not the instructions. I agree that would be annoying otherwise.

> I see where you're coming from, but in this scenario you wouldn't get JS highlighting on the website then. I'm happy to, if we think it's worth the tradeoff. 

I think we can re-create js highlighting without codemirror v easily, (render-codeblock-noncmjs >` {{ transform.Highlight .Inner "javascript"}}` so then I guess it's a vote on which is the most common/default desired case: code mirror or flat code.

Let's take a vote? @Dedekind561 @illicitonion @gregdyke ?

**Review by SallyMcGrath — APPROVED** [review]
LGTM

**Comment by 40thieves** [comment]
> Literally just "interactive code"

Ah I see, now done.

> I think we can re-create js highlighting without codemirror v easily,

I'm persuaded by your argument, so I've updated it in the code.

**Comment by 40thieves** [comment]
I also bumped up the font size to 1rem, matching the default code block:

Screenshot 2023-05-24 at 20 40 17

File: layouts/partials/foot.html [hunk 1]
```diff
@@ -34,5 +34,7 @@
 </footer>
 {{ $scripts := resources.Get "scripts/app.js" | resources.Minify }}
 {{ $menuButton := resources.Get "scripts/random-emoji.js" | resources.Minify }}
+{{ $codemirror := resources.Get "scripts/cm6.js" | js.Build }}
```
40thieves, line 37: I was about to merge, but does this need to be minified? @SallyMcGrath

40thieves: Experimenting with it, seems like the answer is yes so I'll add that too

PR #17, opened 2023-05-25
Rewrite js1 week 2 prep content

Dedekind561:
## What does this change?

Module: JS1
Week(s): 2

## Checklist


content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+```js {linenos=table,hl_lines=["6"],linenostart=1}
+function convertToPercentage() {
+  const num = 0.5;
+  const percentage = `${num * 100}%`;
+}
+const result = convertToPercentage(0.3);
+console.log(result);
+```
+
+If our function behaves as intended, we expect the output to be `"30%"`.
```
illicitonion, line 178: Do we want to say this given the exercise below? Maybe _after_ the exercise?

Also, can we differentiate between what `result` evaluates to (`"30%"`) and what's printed to the console (`30%` without quotes)

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+{{<note type="question" title="exercise">}}
+Given the code above, try predicting what `result` will evaluate to.
+
+{{</note>}}
+
+### Undefined values
+
+When we log result's value to the terminal we get a new data type - `undefined`.
```
illicitonion, line 187: This isn't actually _new_, it was introduced in week one - we should maybe frame this as reinforcement rather than introduction?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+{{</note>}}
+
+### Undefined values
+
+When we log result's value to the terminal we get a new data type - `undefined`.  
+`undefined` is a data type used to represent the absence of a value.  
+We need to specify the value want to output from our function. If we don't state what our function should output then the default output will be `undefined`.
+
+### Setting an output
```
illicitonion, line 191: It's reference the idea of expressions evaluating to values again here - it was heavily referenced in week 1, I think we should be reinforcing that terminology and mental model here.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+  const percentage = `${num * 100}%`;
+  return percentage;
+}
+```
+
+Now that our function contains a `return` statement we can re-run our script to check the output.
+
+### Checking with different inputs
+
+Now we have a function declaration we can check if our function works for any number. We can call `convertToPercentage` with different arguments and check the outputs we get each time.
```
illicitonion, line 211: Are you intentionally using the word "arguments" here? You've previously called this "input" - if we're going to introduce the term, we may want a definition block equating "arguments" to "input"?

illicitonion, line 211:
```suggestion
Now we have a function declaration we can check if our function works for any number. We can call `convertToPercentage` with different arguments and check the return value we get each time.
```

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+To understand how our function currently behaves we must start building an understanding of how our code is executed. We might currently have some ideas of how code is executed in our program:
+
+```mermaid
+mindmap
+Root[How is a piece of code executed?]
+    B[Lines of code are executed from top to bottom]
+    C[Variable declarations mean data is stored in memory]
+    D[Function call means running the instructions inside the function]
+    E[When returning from a function we get an output]
```
illicitonion, line 248: Let's rephrase this to not use the word "output"

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+### ➡️ Current line of code
+
+We can make use of execution diagrams to visualise how a program is executed by a computer.
+
+Firstly, we need a way to track the **line of code that is currently executing.** In the diagrams below, we use an arrow to indicate which line of code is currently being executed. Each line is an instruction to the computer that will produce some effect. For example, a line of code with a variable declaration means "store a new variable with a particular value".
+
+Let's consider an example:
+
+<iframe width="800" height="300" frameborder="0" src="https://pythontutor.com/iframe-embed.html#code=const%20name%20%3D%20'Gayle'%3B&codeDivHeight=400&codeDivWidth=350&cumulative=false&curInstr=0&heapPrimitives=nevernest&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false"> </iframe>
```
illicitonion, line 259: Can we trim this _above_ the "Move and hide objects" line?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+In the example above, we have a single variable declaration:
+
+```js
+const name = "Gayle";
+```
+
+which means store a string `"Gayle"` in memory.
+
+### 🖼️ Storing things in memory
```
illicitonion, line 269: I think before introducing the idea of frames, it would be useful to show an example of what a trainee may think would work that doesn't, which motivates the understanding?

We could show an example of `console.log(percentage)` after the two outputs above, and have them guess what it prints? Then go into this explanation by way of explaining why it doesn't work like they'd expect?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+So far we've seen 2 things:
+
+- lines of code are executed from top to bottom
+- we keep track of variables stored in memory using a global frame
+
+However, we need to consider what happens when we declare and call a function in our program.
+Let's consider another example:
+
+<iframe width="850" height="375" frameborder="0" src="https://pythontutor.com/iframe-embed.html#code=const%20firstName%20%3D%20'German'%3B%0Aconst%20lastName%20%3D%20'Bencii'%3B%0A%0Afunction%20makeGreeting%28%29%20%7B%0A%20%20%20%20const%20greeting%20%3D%20%60Hello%20there%60%3B%20%0A%7D%0A%0AmakeGreeting%28%29%3B&codeDivHeight=400&codeDivWidth=350&cumulative=false&curInstr=5&heapPrimitives=nevernest&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false"> </iframe>
```
illicitonion, line 290: In this example in the iFrame, the first thing isn't actually the function declaration, it's the `firstName` variable. I think we should edit the example so it actually _does_ start with the function declaration (we can come back at some later time to explain "Surprise! Functions can actually be defined out of order!"), but I think that's too much noise for still understanding the idea of scope at all.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+The first thing in the program is the function declaration for `makeGreeting`.
+
+`makeGreeting` is stored in the Global frame as an **object**. We don't need to explore objects as this point - all we need to understand at this stage is that objects are a data type in JavaScript. We will return to objects in more depth later on in the course.
```
illicitonion, line 294: Why don't we just say it's stored as a `function` type?

That's how we've introduced `Math.round` before, and it's what node will print if you inspect it:

```console
> function beep() {}
undefined
> beep
[Function: beep]
> typeof beep
'function'
```

introducing the idea of an object seems like unhelpful noise at this point?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+On line 11 we have a function call:
+
+```js {linenos=table,hl_lines=["7"],linenostart=1}
+function makeGreeting() {
+  const greeting = "Hello there!";
+}
+const firstName = "German";
+const lastName = "Bencci";
+
+makeGreeting();
```
illicitonion, line 305: Let's also add one line _after_ this to show what happens _after_ `makeGreeting` is called (i.e. that the program resumes after the call)

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+{{<note type="activity" title="exercise">}}
+
+Earlier we noted that when the computer executes the code above, we get the following output:
+
+```bash
+50%
+50%
+```
+
+Given the code above, try using the visualisation diagrams to "play computer" with this example. Try stepping through each line of code to explain why the program is giving this output
```
illicitonion, line 356: We haven't defined what playing computer is before telling our trainees to "play computer" - I think we should add an explicit definition above, along the lines of "read through the program line be line, saying out loud what each value we see is, and what the computer will do next"

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+  const num = 0.5;
+  const percentage = `${num * 100}%`;
+  return percentage;
+}
+
+const output1 = convertToPercentage(0.5);
+const output2 = convertToPercentage(0.23);
+```
+
+when the function `convertToPercentage` is called, the variable `num` gets recreated in the `convertToPercentage` frame and set to `0.5` every time we call the function. Instead of `num` being reset every time the function is called, we want to make use of the input that is passed to the function `convertToPercentage` every time it is called.
```
illicitonion, line 375: We may want to reference how we passed a value `10.3` to `Math.round` and it did something based on that actual value?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
 backlog_filter= 'Week 2'
 +++
 
+## Learning objectives
 
+- Label and describe the syntactic features of a function declaration
+- Identify a call site
+- Differentiate between a function declaration and a function invocation
```
illicitonion, line 15: I feel like there's a piece at the end wrapping up and recapping these things

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Propose a function definition to satisfy some problem requirements
+- Given a short program, determine which variable names are _bad_ choices
```
illicitonion, line 17: We definitely haven't touched on this this week.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Give examples of when the `undefined` data type may occur in a JavaScript program
```
illicitonion, line 18: I'm not sure this is achieved this week yet.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Construct simple expressions to represent a set of operations
+- Define a _parameter_
+- Identify the return value a call site given its function declaration
+- Identify the value assigned to a parameter when a function is invoked
```
illicitonion, line 22: I think we could do with a sentence or two explicitly explaining this at the end, and contrasting the terms "parameter" and "argument".

In particular, I don't think we've explained very clearly yet that "a parameter is a variable, which only lives for the scope of the function call, whose value is defined by the caller" - we've hinted at all of these things, but not solidified them all together.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Identify the type of a return value given some call site
+- Given a short program, state the number of times a given function is called
+- Differentiate between parameters and arguments
```
illicitonion, line 25: I don't think we've done this yet.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Identify a call site
+- Differentiate between a function declaration and a function invocation
+- Propose a function definition to satisfy some problem requirements
+- Given a short program, determine which variable names are _bad_ choices
+- Give examples of when the `undefined` data type may occur in a JavaScript program
+- Construct simple expressions to represent a set of operations
+- Define a _parameter_
+- Identify the return value a call site given its function declaration
+- Identify the value assigned to a parameter when a function is invoked
+- Identify the type of a return value given some call site
```
illicitonion, line 23: I don't think we've really covered this?

illicitonion, line 23: I don't think we've really covered this?

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Given a short program, state the number of times a given function is called
+- Differentiate between parameters and arguments
+- Invoke a given function with appropriate arguments to produce some target output
```
illicitonion, line 26: We've only covered a single parameter/argument so far. I'd either make this explicitly singular, or show an example with multiple parameters.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Describe how to log the output of a function
+- Analyse each line in a given program and state what the line does when the program executes
+- Explain the difference between returning a value and logging a value
+- Assess and explain whether a function can reference a given variable or not
+- Explain what is meant by the term "implementation opacity"
+- Explain how to reuse some code to perform a specific task
+- Propose clear and meaningful parameter names for a function
```
illicitonion, line 33: I don't think we've talked about this enough yet.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
 backlog_filter= 'Week 2'
 +++
 
+## Learning objectives
 
+- Label and describe the syntactic features of a function declaration
+- Identify a call site
+- Differentiate between a function declaration and a function invocation
+- Propose a function definition to satisfy some problem requirements
```
illicitonion, line 16: I'm not sure this is achieved in this week so far.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+- Differentiate between parameters and arguments
+- Invoke a given function with appropriate arguments to produce some target output
+- Describe how to log the output of a function
+- Analyse each line in a given program and state what the line does when the program executes
+- Explain the difference between returning a value and logging a value
+- Assess and explain whether a function can reference a given variable or not
+- Explain what is meant by the term "implementation opacity"
+- Explain how to reuse some code to perform a specific task
+- Propose clear and meaningful parameter names for a function
+- Given a problem, suggest clear semantic names for a function designed to solve the given problem
```
illicitonion, line 34: Nor this.

illicitonion, line 34: Nor this.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+b) perimeter of the rectangle
+
+{{</note>}}
+
+{{<note type="exercise" title="Exercise 1.2">}}
+
+The following variable represents the cost of something in pence:
+
+```js
+price; // could have a value like "130p"
```
illicitonion, line 114: This feels like a huge jump and requires working out a bunch of string indexing and stuff?

I'm imagining this looking something like (possibly with some number conversions instead):

```js
function toPounds(penceString) {
  // Assume the string ended with a `p`.
  const penceStringWithoutTrailingP = penceString.substring(0, penceString.length - 1);
  const paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, '0');
  const pounds = paddedPenceNumberString.substring(0, paddedPenceNumberString.length - 2);
  const pence = paddedPenceNumberString.substring(paddedPenceNumberString.length - 2).padEnd(2, '0');
  return `£${pounds}.${pence}`
}
```

Which is a _lot_ of new concepts.

Do you perhaps mean for the values to be numbers instead of strings? Converting `130` to `£1.30` seems a lot more reasonable than converting `"130p"`, with the knowledge from week 1...

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+To specify a function's output, we write a `return` statement.
+
+If your function call is like a question, the return statement is the answer. It's what comes back.
+
+We can add a return statement to the function as follows:
+
+```js {linenos=table,hl_lines=["4"],linenostart=1}
+function convertToPercentage() {
+  const num = 0.5;
+  const percentage = `${num * 100}%`;
```
illicitonion, line 202: I think at some point in this week (doesn't have to be right here, but at some point) we should show an example and make a point of making clear the return value of a function can be any expression, and doesn't need to specifically be a variable.

i.e. that we could rewrite this function as:
```js
function convertToPercentage() {
  const num = 0.5;
  return `${num * 100}%`;
}
```

and that it will work the same.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+### Useful expressions
+
+It is often helpful to solve a problem in one specific instance before doing it for all cases.
+
+In programming, we always try the simplest thing first. Let’s consider how to convert just one number in to a percentage.
+Let’s consider the variable declaration below:
+
+```js {linenos=table,linenostart=1}
+const num = 0.5;
```
illicitonion, line 79: Throughout, I think we should use the variable name `decimal` or `decimalNumber` or similar here, rather than `num`. "num" could easily be a % just as a number without a % after (i.e. `const num = 100;` could be meant to become `"100%"`). Let's make an example of how we can make that more clear though naming :)

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+when the function `convertToPercentage` is called, the variable `num` gets recreated in the `convertToPercentage` frame and set to `0.5` every time we call the function. Instead of `num` being reset every time the function is called, we want to make use of the input that is passed to the function `convertToPercentage` every time it is called.
+
+### Parametrising a function
+
+To make this function reusable for any number, we need to handle inputs. We do this using a **parameter**.
+
+🔑 A parameter is a variable that enables us to reference the input to a function.
+
+We can add a parameter `num` to our function, as with the execution diagram below:
```
illicitonion, line 383: We should make clear that we're _replacing_ the variable definition with a parameter with the same name. It could be easy to miss that up-lifting is happening. This would be a great place to add some explicit text along the lines of "A parameter is a variable whose value is defined by the caller - it's a variable just like the `num` variable was before, but now that it's a parameter we don't assign it a value in the function, the caller gives a value which will be assigned to it".

illicitonion (commented):
This is looking really good! As with w1 I've opened a PR against this branch with some edits https://github.com/CodeYourFuture/curriculum/pull/160 as well as a bunch of comments below - happy to discuss anything!

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+
+```js
+price; // could have a value like "130p"
+```
+
+Write an expression for the price in pounds. The price in pounds should start with a "£" symbol. The price should be to 2 decimal places.
+
+{{</note>}}
+
+### ♻️ Wrapping up functionality
```
carl0FF, line 121: I have the feeling that we are jumping into defining functions without having properly shown what a program is (because all we've seen up to now are one-liners). I think it would be valuable to explain/show that (1) a program is a set of instructions that are executed in sequence, and that (2) we go through a series of intermediate states until (3) we reach the desired result. That could help reinforce the idea of why functions are a useful tool.

I think the code visualizer is an awesome tool for understanding what happens in a program and embedding it in the content is a brilliant idea. What if, before jumping into functions, we show a simple program running in the code visualizer?

For instance:
```javascript
// Let's calculate what an amount of seconds is in days, hours, minutes and seconds
const totalSeconds = 340368;

const remainingSeconds = totalSeconds % 60;
const totalMinutes = (totalSeconds - remainingSeconds) / 60;
const remainingMinutes = totalMinutes % 60;
const totalHours = (totalMinutes - remainingMinutes) / 60;
const remainingHours = totalHours % 24;
const days = (totalHours - remainingHours) / 24;

const result = `days: ${days}, hours: ${remainingHours}, minutes: ${remainingMinutes}, seconds: ${remainingSeconds}`;
console.log(result);
```
By running step by step through this program with the code visualizer, the trainee will see as intermediate results are being calculated and stored to, in sequence, reach the desired result. This is just an example, maybe we could come up with a simpler piece of code, because we don't necessarily want the trainee trying to understand this concrete algorithm, we just want the trainee to understand what a program is, and how it's executed.

This change would require moving the "Current line of code" and "Storing things in memory" sections up, though.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+    A[0.23] --> B{convertToPercentage}
+    B --> C[23%]
+```
+
+This time we have a difference in that we have defined a parameter `num` in the function declaration inside parentheses after the function name `convertToPercentage`. In our current mental model, a function call means going to `convertToPercentage` and running the code inside the function.
+
+<iframe width="800" height="350" frameborder="0" src="https://pythontutor.com/iframe-embed.html#code=function%20convertToPercentage%28num%29%20%7B%0A%20%20const%20percentage%20%3D%20%60%24%7Bnum%20*%20100%7D%25%60%3B%0A%20%20return%20percentage%3B%0A%7D%0A%0Aconst%20output1%20%3D%20convertToPercentage%280.5%29%3B%0Aconst%20output2%20%3D%20convertToPercentage%280.3%29%3B&codeDivHeight=400&codeDivWidth=350&cumulative=false&curInstr=0&heapPrimitives=nevernest&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false"> </iframe>
+
+{{<note type="activity" title="exercise">}}
+Use the visualiation digram to see what happens when the code is above is executed. Pay special attention the lines when `convertToPercentage` is called.
```
carl0FF, line 408:
```suggestion
Use the visualiation diagram to see what happens when the code above is executed. Pay special attention the lines where `convertToPercentage` is called.
```

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+{{</note>}}
+
+{{<note type="exercise" title="exercise">}}
+Use the visualiation digram to see what happens when the code is above is executed. Pay special attention the lines when `convertToPercentage` is called.
+{{</note>}}
```
carl0FF, line 413: Duplicate note

content/js1/sprints/2/prep/index.md
carl0FF: Additionally I'd say we'd need to display the updated code in a visually consistent way, so I'd include  a code block of the likes:

```js {linenos=table,hl_lines=[7] ,linenostart=1}
function convertToPercentage(num) {   // now num is a parameter of convertToPercentage
  const percentage = `${num * 100}%`;
  return percentage;
}

const output1 = convertToPercentage(0.5);
const output2 = convertToPercentage(0.23);
```

I know the updated code is in the code visualizer a bit below, but I'm afraid the change can be easily missed by the different visualization, which may be distracting.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+const output2 = convertToPercentage(0.23);
+```
+
+when the function `convertToPercentage` is called, the variable `num` gets recreated in the `convertToPercentage` frame and set to `0.5` every time we call the function. Instead of `num` being reset every time the function is called, we want to make use of the input that is passed to the function `convertToPercentage` every time it is called.
+
+### Parametrising a function
+
+To make this function reusable for any number, we need to handle inputs. We do this using a **parameter**.
+
+🔑 A parameter is a variable that enables us to reference the input to a function.
```
carl0FF, line 381: Maybe a definition note here for emphasis?

carl0FF: Additionally, this is probably the point at which we want to introduce the term "argument" as synonym of parameter.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+We could visualise what happens when `makeGreeting` is passed a specific argument:
+
+```mermaid
+
+flowchart LR
+    A[0.23] --> B{convertToPercentage}
+    B --> C[23%]
+```
+
+This time we have a difference in that we have defined a parameter `num` in the function declaration inside parentheses after the function name `convertToPercentage`. In our current mental model, a function call means going to `convertToPercentage` and running the code inside the function.
```
carl0FF, line 403: What about the fact that functions can have more than one argument? Once we've established the basic mental model for arguments, it seems to me like the cognitive load of showing that arguments can be multiple at this point is minimal. We could revisit the examples of area and perimeter to show how to use multiple arguments.

carl0FF (commented):
HUGE work in this and the previous PR, I love how concepts are being introduced with just the right amount of detail to keep the focus on what matters at this stage 👏

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+As variables are declared, we need a way to keep track of the memory that is stored as our program executes.
+In our diagrams, we keep track of memory using a _frame_.
+
+For now, we can think of a frame as a space in our diagram that represents the variables stored in memory in some part of our code.
+
+The _global frame_ is always the first frame that gets created when our program starts executing. We can think of the _global frame_ as a location for variables created outside of any function. When we run the variable declaration above, we get a string `"Gayle"` stored in the Global frame:
+
+![global-frame](global-frame.png)
+
+### 🏃 Executing functions
```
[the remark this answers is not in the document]
40thieves: Yeah I would actually vote to rename this section to "Calling functions" as I think that's the more accurate term?

I wouldn't be against a note to say that "execute" is also common term. EDIT: "run" is probably also worth noting.

content/js1/sprints/2/prep/index.md
40thieves: What would you put on that line? A "dummy" log? I feel that would just be distracting?

40thieves (commented):
Sorry for a real late comments here, but agree with Daniel & Carlos that this is looking great!!

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
 
+## Learning objectives
 
+- Label and describe the syntactic features of a function declaration
+- Identify a call site
+- Differentiate between a function declaration and a function invocation
+- Propose a function definition to satisfy some problem requirements
+- Given a short program, determine which variable names are _bad_ choices
+- Give examples of when the `undefined` data type may occur in a JavaScript program
+- Construct simple expressions to represent a set of operations
```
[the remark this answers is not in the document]
Dedekind561: Maybe "represent a calculation" works better?

content/js1/sprints/2/prep/index.md
Dedekind561: I think I meant to put 130 here yeah.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+```
+
+When we execute the code we _want_ to log the target output for each input `0.5` and `0.3`:
+
+```bash
+50%
+23%
+```
+
+However, _given the function's current implementation_, we get the following logs:
```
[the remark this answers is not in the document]
Dedekind561: I think we should assume they're aware of the term by this point: however, if it transpires they're not familiar with it, then we can go back and update some of the earlier content to ensure it's covered.

content/js1/sprints/2/prep/index.md
```diff
@@ -8,4 +8,406 @@ backlog= 'Module-JS1'
...
+Previously, we saw how to use some functions to perform a given task. Here are some examples:
+
+```js {linenos=table,linenostart=1}
+console.log("hello world"); // prints "hello world" to the console
+Math.round(3.141); // will output 3
+```
+
+All these expressions are _function calls_: we’re passing input to the functions, `console.log` or `Math.round` to use their functionality. In the case of `Math.round` and `console.log` these functions have been created and made available to use whenever we start running a program with NodeJS.
+
+However, there is so no such pre-built function that allows us to convert any number to a percentage. So we're going to create a function called `convertToPercentage` with the following properties:
```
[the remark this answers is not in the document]
Dedekind561: I define _specification_ in more detail later on - around week 4. I'm going to change this to _requirements_ - a term they should be familiar with.

content/js1/sprints/2/prep/index.md
Dedekind561: I agree the cognitive load is minimal - and that's why I've not left it in here. There will be exercises where they'll encounter this concept and if their mental model is sufficiently good enough then they'll be able to pick up the idea of 2 parameters without us explaining it here.
I'm aiming to build a minimal viable mental model here.

content/js1/sprints/2/prep/index.md
Dedekind561: I agree it is noise -  if it wasn't for the visualiser I would omit the reference to the **Objects** section. If we make our own widget then we can remove the reference to objects here. I think we should just keep it for now so that it is consistent with the widget.

content/js1/sprints/2/prep/index.md
Dedekind561: Yeah the example below was correct - hoisting is noise at this point.

content/js1/sprints/2/prep/index.md
Dedekind561: I think we should address this learning objective in depth in one of the coursework issues

content/js1/sprints/2/prep/index.md
Dedekind561: Going to leave as is for the time being, but I agree we want to emphasise the resumption of the program after the call.  Need to think of a better example really.

Dedekind561:
@SallyMcGrath this is ready for your copy edits :)

SallyMcGrath (approved):
Note, as this is a final pass, I've made my edits assuming:

1. tooltips (now merged into main) exist
2. but not yet revised block paths -so when #246 is merged these pages need to also be updated linking that PR now

I also turned on footnotes, and I don't normally add features in content but in this case this is ready to go so I don't want to pointlessly block it

🧐 Note: throughout these blocks, `num` and `decimalNumber` were used interchangeably, but this was not referred to in the text so I presumed it was a mistake. I have changed them all to decimalNumber and also updated the python tutor iframes to match. Hopefully this was the right decision and hopefully I've caught them all.

I removed some learning objectives that I couldn't find much evidence for. You might want to put them back and, if so, do, but make it clearer to what they apply?

Dedekind561:
Yeah `decimalNumber` was the preferred variable name: I just missed changing them throughout this copy- thankyou for catching that. I think some of the learning objectives were shifting about between different weeks so it makes sense some of them should go.

PR #19, opened 2023-05-25
Rewrite js1 week 3 prep content

Dedekind561:
## What does this change?

Module:
Week(s):

## Checklist

## Rendered Pages

[Rendered prep page](https://deploy-preview-19--cyf-curriculum.netlify.app/js1/sprints/3/prep/)


content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+```js {title="main.js"}
+if (condition) {
+  // do some code in here
+}
+```
+
+The `if` statement consists of:
+
+1. `if` keyword: this is the start of the `if` statement
+2. `(condition)`: some conditional expression that is wrapped in parentheses.
```
illicitonion, line 465:
```suggestion
2. `(condition)`: some expression that is wrapped in parentheses.
```

I've seen a lot of our trainees think that you can _only_ use expressions containing a comparison in `if` conditions, so write things like `if (someBoolean === true)` rather than just `if (someBoolean)` because they don't grok that the condition can be _any_ expression.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+{{<note type="exercise" title="Exercise 2.1">}}
+Write a program that defines a variable `eleniAge`.
+
+If Eleni is less than 18, the program should print out nothing to the console.
+
+If Eleni is 18 years old or older, the program should print out:
+
+```bash
+Eleni is old enough to drive
+```
```
illicitonion, line 490: This exercise feels a lot more like the "writing very forced code that doesn't really make sense" style of exercise we had before this rewrite.

The instruction "Write a program that defines a variable `eleniAge`" (and then implicitly assume that the trainee has written an assignment with a meaningful value, and the implicit idea that they're going to maybe change that value to experiment) assumes too much contextual knowledge about why we're making this variable, what values are being assigned, and how it should be changed.

I think it'd make more sense to either frame this as "write a function", or as "Here's some code, add to it / fill in the gaps". I think framing it as "write a function" probably makes a lot of sense, given what we've been teaching and where things are placed, but if not, I'd probably provide this script:

```js
const eleniAge = 10;
```

then instruct the trainees to copy that and write their conditional code underneath it.

And then later on have an explicit instruction "Change the value of `eleniAge` to `20` and check your code output what was intended"

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+```bash
+Eleni is not old enough to drive
+```
+
+{{</note>}}
+
+### 🧠 Applying new knowledge
+
+So for `formatAs12HourClock` we said part of the strategy for handling `"23:00"` would involve checking that the `hours` value is less than 12. For this purpose, we can use another comparison operator: `>`.
+`>` will check if the value on the left of the operator is less than the value on the right of the operator.
```
illicitonion, line 507:
```suggestion
`>` will check if the value on the left of the operator is greater than the value on the right of the operator.
```

You may want to switch around the phrasing of this whole section to either be concerning "greater than" in English, or use the `<` operator - mixing the two is probably confusing (and `>` vs `<` is already a key point of confusion in our trainees- let's avoid needing to add implicit negations in people's heads :))

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+So `3 > 12` would evaluate to be `false`, as `3` is not greater than `12`.
+
+So provided we have an expression for the hours from the time, we can write a conditional statement as follows:
```
illicitonion, line 510:
```suggestion
So provided we have an expression for the hours from the time as a number, we can write a conditional statement as follows:
```

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+```js
+if (expressionsForHours < 12) {
+  // do code to format the 12 hours
+}
+```
+
+To complete the logic, we now have to achieve a sub-goal:
+
+> 🎯 Sub-goal: Find an expression for the hours from the `time` input
```
illicitonion, line 520:
```suggestion
> 🎯 Sub-goal: Find an expression for the hours from the `time` input, as a number
```

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+In JavaScript, we can use square bracket notation to access specific characters in the string using the index.
+
+```js
+time[0]; // evaluates to "2"
+time[1]; // evaluates to "3"
+time[2]; // evaluates to ":"
+// etc
+```
+
+However at the moment, square bracket access wil only give a single character.
```
illicitonion, line 544:
```suggestion
However square bracket access will only give a single character.
```

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+So we must use another method to extract _multiple_ characters from the given string.
+
+### ✂️ Extracting a slice
+
+We can use a function called `slice` to extract 1 or more characters from a string.
```
illicitonion, line 549: This is the first time we've introduced a _method_ which has a bound `this` value, as opposed to a function with no `this` - it's _probably_ ok to skip over this, but it may be worth calling out in a note

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+[`slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) is a function that can take 2 arguments: a start index and an end index. `slice` will return a section of the string from the start index up to but not including the start index.
```
illicitonion, line 551:
```suggestion
[`slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) is a function that can take 2 arguments: a start index and an end index. `slice` will return a section of the string from the start index up to but not including the end index.
```

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+```js
+time.slice(0, 2); // will access the characters below
+```
+
+|   index   |   0   |   1   |
+| :-------: | :---: | :---: |
+| character | `"2"` | `"3"` |
+
+So `time.slice(0,2)` will evaluate to `"23"` when the `time` is `"23:00"`. Finally we must convert `"23"` to the number `23`, otherwise we can't compare this value properly. We can use the `+` operator to convert strings to numbers by writing the following:
```
illicitonion, line 565: I'd much rather introduce `parseInt` or the `Number()` constructor rather than `+` as a coercing operator... If someone introduced `+someString` in a PR I was reviewing, I'd ask them to change it out... WDYT?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+A{Check: Are the hours greater than 12?}
+
+```
+
+Now we can format the string using our approach from earlier:
+we’ll need to append `"pm"` to the string expression and subtract 12 from the hours. So we get the following:
+
+```js
+if (+time.slice(0, 2) > 12) {
+  return `${time.slice(0, 2) - 12}:00 pm`;
```
illicitonion, line 604: I don't think until now we've introduced the idea that a `return` statement _also stops function execution_. So I don't think it's obvious _what to do with this new `if` statement.

I think we either need to introduce here the idea that you can early-`return`, or introduce the idea of an `else` statement.

I think the leap, particularly if this is "do on your own at home" prep work, of what to do with this `if` statement in relation to the existing code, is too big a leap, without one of those two concepts.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+However, at the moment, we’re making use of the same expression twice: `+time.slice(0,2)`. This means we’re calling the function `slice` twice. Additionally, expressions embedded inside curly braces and parentheses can often be difficult to read. In this situation it makes sense to label the recurring expression so we can reuse wherever it we need to in our code.
+
+Let’s create a variable called `hours` and assign to it our expression.
+
+```js
+function formatAs12HourClock(time) {
+  const hours = +time.slice(0, 2);
+
+  if (hours > 12) {
+    return `${hours - 12}:00 pm`;
```
illicitonion, line 635: We should probably also also have a section where we talk about how this code is buggy because it assumes there are no minutes, and prompt the trainees to add a new assertion and fix that themselves. They already have all of the building-blocks they need to do this, so it should just be a matter of re-applying those same concepts around slicing.

I'd be tempted to also set a piece of coursework around considering the classes of input, and interesting edge cases, they would maybe want to test. Let's note in that exercise also that there are two major ways of identifying these classes/edge-cases: "Domain-driven" (i.e. "morning vs afternoon", "midnight and midday", etc), and "Implementation-driven" (e.g. remembering that we need to include the correct minutes isn't necessarily a core _domain_ problem, but we _can_ identify it as an interesting case to test based on knowing what our implementation looks like - we're separating the hours from minutes, so may need to pay extra attention to the idea that we may do something wrong with the minutes). This would be a great discussion for TAs to facilitate on the Saturday.

illicitonion, line 635: We should probably also also have a section where we talk about how this code is buggy because it assumes there are no minutes, and prompt the trainees to add a new assertion and fix that themselves. They already have all of the building-blocks they need to do this, so it should just be a matter of re-applying those same concepts around slicing.

I'd be tempted to also set a piece of coursework around considering the classes of input, and interesting edge cases, they would maybe want to test. Let's note in that exercise also that there are two major ways of identifying these classes/edge-cases: "Domain-driven" (i.e. "morning vs afternoon", "midnight and midday", etc), and "Implementation-driven" (e.g. remembering that we need to include the correct minutes isn't necessarily a core _domain_ problem, but we _can_ identify it as an interesting case to test based on knowing what our implementation looks like - we're separating the hours from minutes, so may need to pay extra attention to the idea that we may do something wrong with the minutes). This would be a great discussion for TAs to facilitate on the Saturday.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
 backlog_filter= 'Week 3'
 +++
 
+## Pre-requisites
 
+## Learning objectives
+
+- Write an assertion for a function that implements a solution to a given problem
```
illicitonion, line 15: I worry that we may be slightly over-tying assertions to _functions_ - do we maybe want to also introduce an example of an assertion that isn't related to a function, just to show it can be done?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Construct a flow diagram that describes how to produce an output from a given input
+- Explain when it is necessary to use a conditional statement
+- Identify the syntactic features of an `if` statement
+- Describe the difference between `console.log` and `console.assert`
+- Evaluate an expression that uses a comparison operator
+- Describe how to convert a string into a number in JavaScript
+- Implement a piece of functionality specified in an assertion
+- Define _block scope_
+- Define _zero-indexing_
+- Given a short program, determine the return value of a call to `slice` before execution
```
illicitonion, line 27:
```suggestion
- Given a short program, determine the return value of a call to `slice` on a string before execution
```

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Given a problem, suggest potential sub-goals required to reach a solution
```
illicitonion, line 28: I think if we have this as a LO (which I think we should), we maybe need to linger on sub-problems a little more in our exposition?

Maybe where we introduce the idea of a sub-problem we can break down our problem into _all_ of its sub-problems/stages, rather than just showing the one example of a sub-problem? Or reflect back on all of the sub-problems at the end?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Determine the output of a program that uses a conditional statement
```
illicitonion, line 29: "Output" feels vague here. Are we talking about what's logged to the console? Let's call this "behaviour" maybe?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Analyse a conditional statement to determine whether or not the statement will be executed
```
illicitonion, line 30: I think this wording is a bit unclear, probably because we're using the phrase "conditional statement" which is quite broad, and 'the statement' refers to a particular sub-component of the statement... Maybe "Analyse an `if` conditional statement to determine whether its body will be executed"?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Predict and explain how conditional statements involving `else`, `else if` and `while` will be executed
```
illicitonion, line 31: We haven't covered `else`, `else if`, or `while`.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Explain why _refactoring_ is necessary in the production of software
```
illicitonion, line 32: I don't think we've introduced the term `refactoring`, and we have mentioned some motivations but could maybe do with consolidating them. Maybe in the "Improving the code" section, we could have a definition block for "refactoring", and re-enforce the motivations we've gathered so far for why we do it?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+- Propose a strategy for solving a problem
+- Write an assertion to check for some behaviour of a given function
+- Identify a pattern between a set of inputs and outputs produced by a given function
```
illicitonion, line 35: I'd add some LOs around identifying classes of input data, and edge-case identification.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+```bash
+Eleni is old enough to drive
+```
+
+{{</note>}}
+
+{{<note type="exercise" title="Exercise 2.2">}}
+
+Modify the program above so that if Eleni is less than 18 the program prints out:
```
illicitonion, line 496: I'd add a prompt here to change the value of the variable to test out both branches of the code, to make sure their code works for both ages.

(I'd also take this opportunity to reinforce "We thought about the problem and identified the key classes of values - "less than 18" and "18 or over" to manually test, and maybe even go on a slight detour to say "18 is a boundary condition - it's probably also a good value for us to test")

illicitonion (commented):
This generally looks great, nicely put together!

Major themes in my comments:
1. Let's introduce more discussion of classes of input data, and edge-cases
2. We should probably have introduced both syntax errors and reference errors in week 1 or 2, and in particular lingered on how to read an error message and take useful information out of it. Maybe also block scope.
3. We're talking about "statements" a lot now but haven't introduced them before or here. We're just talking about them.

illicitonion (commented):
This generally looks great, nicely put together!

Major themes in my comments:
1. Let's introduce more discussion of classes of input data, and edge-cases
2. We should probably have introduced both syntax errors and reference errors in week 1 or 2, and in particular lingered on how to read an error message and take useful information out of it. Maybe also block scope.
3. We're talking about "statements" a lot now but haven't introduced them before or here. We're just talking about them.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+```
+
+So now we've found an expression for the `hours`, we can write:
+
+```js
+if (+time.slice(0, 2) > 12) {
+}
+```
+
+If the time is `"23:00"` then the condition `+time.slice(0,2) > 12` will evaluate to `true` and the code inside the curly braces will be executed.
```
carl0FF, line 587:
```suggestion
If the time is `"23:00"` then the expression `+time.slice(0,2) > 12` will evaluate to `true` and the code inside the curly braces will be executed.
```
A _condition_ should include the `if`

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+Now we can **re-run** our assertions from earlier to check our function behaves as target.
+
+### 🧹 Improving the code
+
+Now the assertions pass: in other words, our function’s current output matches with the target output described in the assertions.
+
+However, at the moment, we’re making use of the same expression twice: `+time.slice(0,2)`. This means we’re calling the function `slice` twice. Additionally, expressions embedded inside curly braces and parentheses can often be difficult to read. In this situation it makes sense to label the recurring expression so we can reuse wherever it we need to in our code.
+
+Let’s create a variable called `hours` and assign to it our expression.
```
carl0FF, line 628:
```suggestion
Let’s create a variable called `hours` and assign to it our expression's [evaluation] result.
```

carl0FF (commented):
Great job 👏 👏 👏

content/js1/sprints/3/prep/index.md
40thieves: I wonder this is also an opportunity to use `console.assert` to validate the code?

content/js1/sprints/3/prep/index.md
40thieves: I still use the mnemonic "the crocodile eats the bigger one" to this day 😂 I do wonder if this would be helpful as a tip?

content/js1/sprints/3/prep/index.md
40thieves: I think this and some of the comments would read a lot cleaner if we changed from looking to find the value, not the expression. Something like:

> Sub-goal: Find the hours value from the `time` input

I think that's also semantically more correct? We're not really interested in the expression itself, just the value that it evaluates to.

content/js1/sprints/3/prep/index.md
40thieves: Daniel: do you mean that this is a method on the instance of the string, instead of a static function like `Math.round`?

40thieves (commented):
Another round of strong agreement with Daniel & Carlos :) Really like the focus on testing/building up a solution!

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+  console.assert(
+    currentOutput === targetOutput,
+    "current output: %s, target output: %s",
+    currentOutput,
+    targetOutput
+  );
+}
+
+{
+  // ❌ this assertion now fails
```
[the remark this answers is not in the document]
illicitonion: I strongly agree that sneaking TDD in is great, and I agree that introducing "failing for the right reason" is good, but I'd lean pretty heavily towards doing so intentionally. Writing an intentionally failing test, or writing a surprisingly failing test are both great introductions to this, but confounding that with "We introduced a SyntaxError (and maybe the first SyntaxError we've ever encountered), oh no" is more likely to make people confused and scared, rather than appreciate the value of an intentionally failing test...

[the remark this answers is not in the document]
40thieves: Yeah that's fair 👍🏻

content/js1/sprints/3/prep/index.md
illicitonion: Yeah, exactly that Ali :)

40thieves: Thanks, I guessed that but wanted to make sure.

I feel like this might be one of those things that trying it out in a lesson (and looking out for confusion) is probably the best way to find out?

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+### 🧠 Applying new knowledge
+
+So for `formatAs12HourClock` we said part of the strategy for handling `"23:00"` would involve checking that the `hours` value is less than 12. For this purpose, we can use another comparison operator: `>`.
+`>` will check if the value on the left of the operator is less than the value on the right of the operator.
+So `3 > 12` would evaluate to be `false`, as `3` is not greater than `12`.
+
+So provided we have an expression for the hours from the time, we can write a conditional statement as follows:
+
+```js
+if (expressionsForHours < 12) {
```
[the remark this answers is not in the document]
Dedekind561: Awesome - I did worry that the variable without a declaration would be confusing. I think the comment is clearer here.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+// evaluates to false
+```
+
+{{<note type="exercise" title="Exercise 1.2">}}
+
+Look at the code below and then try predicting what each expression will evaluate to. You may want to use the Node REPL to help you check your answers. Check out some relevant documentation if you've not seen a particular operator before.
+
+```js {linenos=table,linenostart=1}
+true === false;
+5 == 2 + 4;
```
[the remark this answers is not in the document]
Dedekind561: This was more for them to just explore the behaviour of other comparison operators (including loose equality) - I was expecting them to look it up briefly to check it doesn't behave the same way as strict comparison.

Dedekind561:
> This generally looks great, nicely put together!
> 
> Major themes in my comments:
> 
> 1. Let's introduce more discussion of classes of input data, and edge-cases
> 2. We should probably have introduced both syntax errors and reference errors in week 1 or 2, and in particular lingered on how to read an error message and take useful information out of it. Maybe also block scope.
> 3. We're talking about "statements" a lot now but haven't introduced them before or here. We're just talking about them.

1. I'll address 1. directly in this prep section
2. I went back and added back a section on SyntaxErrors in Week 1 - I forgot to add it an earlier draft even though it's critical we introduce the concept as early as possible 😅 
3. I'm going to go back and add a section on **declarations** for week 1 and then go back and add a section on **statements** for week 2.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+This leads to the question:
+
+> What will `formatAs12HourClock("08:00") === "8:00 am"` evaluate to?
+
+### ✅ ❌ Boolean values
+
+Some values are best represented as strings: any piece of text, a name, address, etc will most likely be stored using the string data type.
+Similarly, the number data type will be used to store any numerical data where we'll need to do standard numerical operations like addition, multiplication etc.
+
+However, if we're comparing values or expressions then we have one of 2 different states: **true** or **false**.
```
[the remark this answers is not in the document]
Dedekind561: I'm hoping that because booleans will appear in different contexts in the coursework trainees will internalise the idea that booleans can be used in a variety of contexts, and not just limited to comparisons.

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+Up to now we've used the log function `console.log`.
+However, we can also write assertions using another function: `console.assert`. Let's look at an example using this function:
+
+```js
+function formatAs12HourClock() {}
+console.assert(formatAs12HourClock("08:00") === "20:00");
+```
+
+The documentation for `console.assert` states that a message will be written to the console if a condition is `false`.
```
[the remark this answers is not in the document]
Dedekind561: Yes definitely! 💯

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+When the code above executes, we now get a log in the console:
+
+```bash
+Assertion failed: current output: undefined, target output: 08:00 am
+```
+
+### 🧰 Implementing the functionality
+
+On line 4, the function is being passed a single argument `"08:00"`. We can parametrise the function and label the input as `time`:
```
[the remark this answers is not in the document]
Dedekind561: I agree with the copy-edit, but don't think we need signatures at this stage

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+When an error is thrown, it means the execution of the program stops at this point and an error report is sent to the user.
+However, we want to do multiple assertions whilst using the same variables.
+
+To reuse variable names, we can make use of a block declaration as follows:
+
+```js
+{
+  StatementList;
+}
+```
```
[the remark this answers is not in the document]
Dedekind561: I think the comment works best here. I agree the `StatementList` looks confusing as it looks like an identifier - I actually took that syntax straight from some documentation that I can't find right now 🤔

content/js1/sprints/3/prep/index.md
Dedekind561: Yeah no problem - let's use `Number()`

content/js1/sprints/3/prep/index.md
Dedekind561: > I think this wording is a bit unclear, probably because we're using the phrase "conditional statement" which is quite broad, and 'the statement' refers to a particular sub-component of the statement... Maybe "Analyse an `if` conditional statement to determine whether its body will be executed"?

How about "Analyse an `if` statement to determine whether its body will be executed"

content/js1/sprints/3/prep/index.md
```diff
@@ -8,4 +8,634 @@ backlog= 'Module-JS1'
...
+
+Look at the code below and then try predicting what each expression will evaluate to. You may want to use the Node REPL to help you check your answers. Check out some relevant documentation if you've not seen a particular operator before.
+
+```js {linenos=table,linenostart=1}
+true === false;
+5 == 2 + 4;
+4 * 5 == "20";
+4 * 5 === "20";
+3 * 2 === 6;
+4 = 4;
```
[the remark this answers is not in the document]
Dedekind561: Good catch: I've added SyntaxErrors to week 1

content/js1/sprints/3/prep/index.md
Dedekind561: I agree - I just copied this over from the old syllabus. I'm going to replace it with something more interesting.

Dedekind561:
Things to improve:
- [] exercises
- [] sub-goal section
- [] give another example of an assertion

Dedekind561:
@SallyMcGrath Thankyou for your copy edits! I'm going to take out the final section on refactoring and set that aside for coursework

SallyMcGrath:
Yeah -  I stopped work on this for Daniel tho

SallyMcGrath:
@Dedekind561 

I merged the new way of calling blocks. When this is ready to go, let me know and I can update the paths. (Or even better you can just do it if you follow the changes!)

Dedekind561:
Sweet, I'll update the paths now and merge this in

PR #22, opened 2023-05-28
module tabs variation

SallyMcGrath:
feedback on fundamentals was that the prep page was too long and the goals should go elsewhere -- I've made this module tabs view to add in this kind of content. I will also set a short piece of coursework to find, read, and answer some basic questions on this content.

### DIFF
--- a/layouts/partials/module-tabs.html
+++ b/layouts/partials/module-tabs.html
@@ -1,0 +1,54 @@
+{{ $parent := .Section }}
+<tab-panels class="c-tabs c-tabs--module">
+  <ul class="c-tabs__list" id="{{ .Title | urlize }}" role="tablist">
+    <li class="c-tabs__item">
+      <a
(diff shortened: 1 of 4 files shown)


layouts/partials/module-tabs.html
```diff
@@ -0,0 +1,54 @@
+{{ $parent := .Section }}
+<tab-panels class="c-tabs c-tabs--module">
```
40thieves, line 2: Out of curiosity why not use the shortcode here?

40thieves:
Screenshot for context:

image

layouts/partials/module-tabs.html
SallyMcGrath: Shortcodes are for markdown

40thieves: Oh of course 🤦🏻‍♂️

Issue #9, opened 2023-05-18
Feature: Highlight sections of a code block


Dedekind561:
Given a code block, 
I want to highlight various parts of the code fence by altering the opacity so to draw a user's attention to a particular section of the code block.

```js
function sum(a,b) {
      return a + b;
}

const result = sum(10,32);
```
Given the code snippet above, I might want to highlight the line with the call site, so the rest of the code block's opacity would need to be lowered so the call site line stands out.

40thieves:
I guess this is a feature request for CM6 then :D

Pretty sure it can be done via [a line decoration](https://codemirror.net/docs/ref/#view.Decoration%5Eline). It's not entirely clear how we'd pass the configured line numbers through to CM6 though.

Dedekind561:
![image](https://github.com/CodeYourFuture/curriculum/assets/25401570/de3e96fb-17ad-4a79-93b5-9c17d64a4f83)

Dedekind561:
Sorry I missed the earlier part of this thread where you shared exactly the same thing lol 🤦

SallyMcGrath:
Also #118

PR #23, opened 2023-05-28
Highlight lines in CM6

40thieves:
## Description

Highlight specific lines of a code block when using CM6. This was raised as a feature request in https://github.com/CodeYourFuture/curriculum/issues/9.

Capture-2023-05-28-231010

I also took the liberty of converting the CM6 code to TypeScript. This is mostly because CM6 itself is written in TS, as is most of the documentation, and so it makes sense to follow that. Frustratingly I still get a type error in VSCode, but it's spurious (at least in browsers that aren't 10+ years old). Hints on how to get rid of this would be welcome.

@Dedekind561 @SallyMcGrath since you requested it :D

### DIFF
--- a/assets/scripts/highlight-lines.ts
+++ b/assets/scripts/highlight-lines.ts
@@ -12,11 +12,9 @@ const highlightedLine = Decoration.line({ class: "hl_line" });
 // Define a ViewPlugin that controls how the highlightedLine Decoration is
 // applied to the view
 const highlightLinesPlugin = ViewPlugin.define(
-  (view) => {
-    return {
-      decorations: Decoration.set(highlightLines(view)),
-    };
-  },
+  (view) => ({
+    decorations: highlightLines(view),
+  }),
   {
     decorations: (viewPlugin) => viewPlugin.decorations,
   }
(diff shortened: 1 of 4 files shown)


assets/scripts/highlight-lines.ts
```diff
@@ -0,0 +1,81 @@
...
+const highlightedLineNumbers: Facet<[number?, number?]> = Facet.define({
+  // Since we only have 1 place this can be configured, we just always take the
+  // first input value
+  combine: ([val]) => val,
+  static: true,
+});
+
+// Define a theme that styles the highlighted lines
+const highlightLinesTheme = EditorView.baseTheme({
+  ".hl_line": { backgroundColor: "hsl(60deg 100% 90% / 0.8)" },
```
40thieves, line 71: I just copied this from the [Hugo docs](https://gohugo.io/content-management/syntax-highlighting/#highlighting-in-code-fences), happy to change it to whatever really. I'm fairly sure that CSS variables would work here too, if wanted.

40thieves:
Ah, forgot to test updates... 

I think I just need to map the changes through updates, so shouldn't be too hard to fix.

Dedekind561:
Awesome, let me know when you're done and I'll review ( although I might be away on holiday 🍹  )

40thieves:
Ok, fixed updates.

It required a bit of re-architecting, as essentially the highlighted lines need to be held in CM6's state mechanism (a `StateField`) so that they update through editor transactions. I think it ended up making it a bit simpler though.

40thieves:
So I think it's ready for review now.

SallyMcGrath (approved):
Great - works great. I will put a theme on at some point soon.

PR #15, opened 2023-05-24
tabs component

SallyMcGrath:
For including stuff like facilitator notes or answers

Written as a web component - maybe we want password locking in the future.

PR #24, opened 2023-05-29
Clean up indentation of tabs shortcode

40thieves:
## Description

Follow up to #15, I was looking to understand the code and ended up tidying up the indentation. I think I also fixed a bug too? I removed a stray `<p>` that shouldn't have been in there.

Sorry for the nitpicky PR!

@SallyMcGrath


SallyMcGrath (approved):
TY!

not sure why my linter didn't run... sorry!

This code is from the Kubernetes docs and I left in the code for grabbing code blocks from bundles because it seemed like it would be useful. However, I would not be opposed to stripping it out to make this block clearer.

40thieves:
> This code is from the Kubernetes docs and I left in the code for grabbing code blocks from bundles because it seemed like it would be useful. However, I would not be opposed to stripping it out to make this block clearer.

Yeah I must admit I mostly got to the point of "this is doing some complicated stuff that I don't think we're using currently" and stopped there. If you think they might be useful down the line, then I defer to your judgement :)

PR #31, opened 2023-06-06
Block call refactor and extension

SallyMcGrath:
developing with PD feedback:

-- create an issue block so issues can be added as content blocks directly not just on the backlog views
-- refactor the pd link strategy so they can just add straight from the front end
-- given these two things I hit a complexity limit for my own brain in the partial and so I restructured it to build a map and added that to the Page Scratch

Now I've done this I can make the learning objectives views simpler as well, and I will add that to this PR if I get it done today


SallyMcGrath:
I'm just gonna merge this as it's blocking stuff!

Will sort PD success objectives later one

40thieves:
Sorry for not getting to this until now, but it LGTM (as far as my understanding of Hugo goes).

Thanks for adding comments, it helps a lot!

PR #41, opened 2023-06-09
PageFind Search

SallyMcGrath:
## What does this change?

Module: all
Week(s): all

## Checklist

## Description

I've added PageFind, https://pagefind.app/

It should be fairly customisable if people want to tweak things.
https://pagefind.app/docs/indexing/

It is a post build step, so if you want to run locally, issue 

`rm -rf public && hugo && npx pagefind --source "public" --serve`

and go to http://localhost:1414/ to see it working. But mainly you can just leave it and serve Hugo in the usual way as it's just a search bar in the header.

@40thieves


illicitonion (approved):
Looks great!

One tweak if it's easy to make (but I suspect may be fiddly because it looks like we're just using a Pagefind component) - it looks like in search results only the title of the page is a link? I instinctively tried to click on the matching text, and also kind of expected the whole `<li>` including its padding to be clickable - is it easy to expand which part is clickable?

SallyMcGrath:
Back from my bath. I also expected this - I bashed on the kbd flow to make the li the focus object. 

We can't easily change the provided UI components, but there's also a full JS API:

https://pagefind.app/docs/api/

So we can prolly:

1. Copy the provided Svelte Components into Hugo... can we just build them with Pipes  or do we need to do something more? (dunno yet)
2. Modify https://github.com/CloudCannon/pagefind/blob/main/pagefind_ui/default/svelte/result.svelte to wrap the inner result in an anchor block
3. namespace our hacked component and call it in much the same way as the canonical one https://github.com/CloudCannon/pagefind/blob/main/pagefind_ui/default/README.md

I am getting on a local train and will pick this up after I've got on the intercity. Don't think it's blocking but it is a good idea.

README.md
```diff
@@ -30,12 +30,21 @@ npm i && hugo server
...
-npm i
+### Site Search
+
+PageFind runs the search. https://pagefind.app/
+It's in the build command on Netlify `hugo && npx pagefind --source "public"`
+If you need to develop on this locally, run:
+
+```zsh
+rm -rf public &&
+hugo && npx pagefind --source "public" --serve
```
40thieves, line 43: Fwiw, this doesn't seem to work for me locally. - the browser fails to find the pagefind assets.

40thieves: Scratch that. I misunderstood how the command works - `npx pagefind` "passes through" the built assets.

bglw:
Sneaking in to mention that [Pagefind's Modular UI](https://www.npmjs.com/package/@pagefind/modular-ui) might be a good fit — the ResultsList component there lets you use custom templating for each result. (Sort of a middle ground between the default UI, and using the JS API directly)

SallyMcGrath:
> Sneaking in to mention that [Pagefind's Modular UI](https://www.npmjs.com/package/@pagefind/modular-ui) might be a good fit — the ResultsList component there lets you use custom templating for each result. (Sort of a middle ground between the default UI, and using the JS API directly)

👋 Hi Liam! Thanks for fixing my issue so amazingly quickly. Come and play - you know you want to! 😉

PR #43, opened 2023-06-10
pass the page object correctly to local blocks

SallyMcGrath:
so we can access the .Section

https://gohugo.io/variables/page/#section-variables-and-methods

Local blocks had disappeared since I refactored the block data
For ex fundamentals/sprints/1/day-plan/#user-stories were blank

This was because the .Page object belongs to the scope, so when passing the scope in this way we need to do $variable.Page.Section - we can't do $variable.Section as $variable is not .Page in this context.

I would like to further refactor this but so many PRs are building up I am loath to do more there. Will leave until Search has been looked at.

### DIFF
--- a/layouts/partials/block-data.html
+++ b/layouts/partials/block-data.html
@@ -12,6 +12,7 @@
 */}}
 {{ $name := .name | default "warning: name is empty" }}
 {{ $src := .src | default "warning: src is empty" }}
+{{/* We need to know what page this block is on */}}
 {{ $p := .page }}
 
 {{/* Initialize the blockData map */}}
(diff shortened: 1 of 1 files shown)


SallyMcGrath:
I'm just going to merge this

40thieves:
Again a bit late, but seems like a good enough workaround in the short-term.

40thieves:
I must admit I'm confused about case sensitivity in Hugo though 🤷🏻

SallyMcGrath:
I am very hazy on this page Page thing...

Also it's too hot and I can't sleep.

PR #50, opened 2023-06-11
swap top link to toc 

### LINKED ISSUE #48: Fundamentals Prep: Skip links are confusing
**Describe the bug**

There's a paperclip link that hovers over the content on the right-hand side. It appears to go to an anchor at the top, and make a button in the top-left appear which says "Skip to main".

I'm not sure why it's there or what it's for - the paperclip icon doesn't have an obvious meaning to me, jumping to top can be useful but it's not clear that's what it does and there are browser-native shortcuts to do so, and showing a "Skip to main" button which appears to unconditionally scroll the content by about 50px doesn't seem super useful?

I'm not sure if these are better tailored to different styles of pages with different layouts, but right now on this page the paperclip is a bit distracting and I'm not really sure why it's there...

Screenshot with paperclip:
Screenshot 2023-06-11 at 00 00 48

Screenshot of "Skip to main" button:
Screenshot 2023-06-11 at 00 01 09

Screenshot after clicking "Skip to main" button:
Screenshot 2023-06-11 at 00 01 17

**Link to content**

https://cyf-curriculum.netlify.app/fundamentals/prep/

SallyMcGrath:
this is a fairly standard accessibility helper I don't think it's a good idea to remove, but def should link to the correct fragment!

fixes #48

### DIFF
--- a/assets/styles/04-components/page-header.scss
+++ b/assets/styles/04-components/page-header.scss
@@ -79,4 +79,8 @@
       display: none;
     }
   }
+
+  &__toc {
+    scroll-margin-top: calc(2 * var(--header-height));
+  }
 }
--- a/assets/styles/04-components/toc.scss
+++ b/assets/styles/04-components/toc.scss
@@ -9,6 +9,14 @@
   max-height: 25vh;
   overflow-y: auto;
 
+  // When scrolling into view, add a bit of spacing above to give room for
+  // sticky header.
+  // Calculate the height of the header, summing:
+  //   45px (rough logo height - magic number since height isn't defined)
+  //   --theme-spacing--1 * 2 (padding around logo)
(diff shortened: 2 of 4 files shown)


illicitonion:
Makes sense!

Looks good, but currently appears to scroll to the middle of the TOC rather than the top? Any idea why?

This is what I see after scrolling down and clicking the link:

Screenshot 2023-06-11 at 11 01 08

SallyMcGrath:
Hrm. It's because there's a sticky header. The top of the toc is at the viewport now. I can fix that but it's a bit fiddly - let's merge this and open a style ticket. (this is actual function)

illicitonion:
I think this is obsoleted by https://github.com/CodeYourFuture/curriculum/pull/87?

Issue #59, opened 2023-06-11
"No response" shows up from issue template responses


illicitonion:
**Describe the bug**

https://cyf-curriculum.netlify.app/fundamentals/sprints/1/backlog/ first exercise has "No response" output from issue template

<img width="773" alt="Screenshot 2023-06-11 at 13 43 01" src="https://github.com/CodeYourFuture/curriculum/assets/1131704/d3da7739-29ca-4cf8-b236-b2d215e765c3">

We should edit out any "No response" sections from GitHub issues. In this instance, I edited it in https://github.com/CodeYourFuture/Course-Fundamentals/issues/8 but it doesn't seem to update on the site. Not sure if there's some caching going on we need to flush...

SallyMcGrath:
It will update - the cache is quite long - maybe 24h.

40thieves:
To me it seems like we should fail the build if we get a bad response from Github? It may not be obvious if or where content is missing in this scenario - and we might not trigger a new build to (potentially) fix it for days/weeks.

However I recognise that might not be possible currently. I assume that throwing an error (or whatever Go's equivalent is) would fail the build? Can this be done from within a template? Would we need to migrate to a Hugo module (I think that's what they're called? I now can't find the relevant docs page 🙄) for this to work?

illicitonion:
> if we get a bad response from Github? It may not be obvious if or where content is missing in this scenario

Just to make sure we're talking about the same thing, the issue isn't that GitHub is failing a request, it's that when someone crated an issue, they did so using a form template, and if you don't answer one of the form template's questions, GitHub still includes the heading for the field, but says _No response_ in the issue.

So content won't be _missing_, it's just a bit messy/misleading on the page.

---

We could search for this in responses from GitHub and fail the build to tell the person doing the build "You should go edit the GitHub issue text and try again" - I just made the edit locally to do this, and got the following list of included issues which have `_No response_` in their body:

<details>
<summary>(118 issues)</summary>

https://github.com/CodeYourFuture/Induction/issues/1
https://github.com/CodeYourFuture/Induction/issues/2
https://github.com/CodeYourFuture/Module-HTML-CSS/issues/1
https://github.com/CodeYourFuture/Module-HTML-CSS/issues/10
https://github.com/CodeYourFuture/Module-HTML-CSS/issues/11
https://github.com/CodeYourFuture/Module-HTML-CSS/issues/12
[...]

PR #89, opened 2023-06-12
Rework GitHub API interactions

illicitonion:
* Disable caching - we want every build to get up to date content for issues, so that if people make clarifications or fixes to issues, we pick them up.
* Allow specifying a GitHub access token when using the GitHub API - otherwise we will hit rate limits (due to a lack of inter-build caching). Unfortunately it's not trivial to make _all_ API requests use auth from one simple function - I'll work out how to simplify this in the future (probably by upstreaming support to Hugo to specify an environment variable to use as a bearer token). For now, though, we'll just copy-and-paste the auth header code.
* Stop ignoring errors on fetch - fetch errors lead to missing content and incorrect builds. Now that we're nearing production, we need to fix these if they come up, rather than ignoring them.
* Warn if fetched issues contain `_No response` placeholders - those issues should be edited in GitHub issues. When we fix them all, I'll upgrade these to errors, to avoid slipping.

Note: I also created the empty repo https://github.com/CodeYourFuture/Module-Final-Projects so that the fetch to /issues of it wouldn't fail, because that was causing build errors with this new not-ignoring-errors way. Hurrah for discovering issues 🎉 

Relates to #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,11 @@ brew install hugo
 
 ### To run locally
 
+You'll need to get a fine-grained GitHub API token which allows read-only access to all public CYF repos from [this page](https://github.com/settings/tokens?type=beta).
+
 ```bash
-npm i && hugo server
+npm i
+CYF_CURRICULUM_GITHUB_BEARER_TOKEN=your_github_api_token hugo server
 ```
 
 ### To create a new module
--- a/config.toml
+++ b/config.toml
@@ -4,7 +4,6 @@ title = "CodeYourFuture Curriculum"
 languageLang = "en"
 sectionPagesMenu = "main"
 enableEmoji = true
-ignoreErrors = ["error-remote-getjson"]
 
 [menus]
     [[menus.secondary]]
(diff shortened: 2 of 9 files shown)


SallyMcGrath (commented):
I moved settings into block-data to clear up these layouts a bit as they are getting incomprehensible and I did want these to be somewhat digestible to trainees at some point.

 I'm also now using block-data outside of a block. So I think maybe the file structure should be more like:

all the blocks in a block folder, so block/readme, block/issue

blockData to be renamed... getData? and this 

```
{{ $headers := (dict) }}
{{ if ne (os.Getenv "CYF_CURRICULUM_GITHUB_BEARER_TOKEN") "" }}
{{ $headers = merge $headers (dict "Authorization" (printf "Bearer %s" (getenv "CYF_CURRICULUM_GITHUB_BEARER_TOKEN"))) }}
{{ end }}
```

to be merged into that map? What do you think?

SallyMcGrath:
Note: I have discussed this with Daniel on Slack and I will take this PR forward and complete it (ideally today!)

illicitonion:
Closed in favour of #113 - thanks @SallyMcGrath!

PR #110, opened 2023-06-14
Revise TOC view look and feel

SallyMcGrath:
- replace paperclip with bookmark
- dump old sketchy day plan look
- tweak everything to "go" a bit more (will obviously polish this much more this week anyway)


40thieves:
Thanks, that looks cleaner.

Still not entirely convinced by the bookmark icon, but I think at this point some (pseudo) user testing would be the most productive

PR #113, opened 2023-06-16
Rework GitHub API interactions (moved from fork)

SallyMcGrath:
I can't push to Daniel's fork so I've checked out the pr and opened a new one to main...

That is almost certainly the wrong thing to do! Don't look at me; I'm monstrous.

This does everything #89 does but just once, in the data.html partial where we construct the resource already. 

It's worth noting adding auth like this raises the barrier to entry, as to locally build you now need a github token (and I guess to be part of the CYF org). It would be nice to extend this at some point to a toggle so interested parties can build without these steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,11 @@ brew install hugo
 
 ### To run locally
 
+You'll need to get a fine-grained GitHub API token which allows read-only access to all public CYF repos from [this page](https://github.com/settings/tokens?type=beta).
+
 ```bash
-npm i && hugo server
+npm i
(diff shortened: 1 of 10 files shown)


illicitonion:
> to locally build you now need a github token (and I guess to be part of the CYF org)

Any old GitHub token will do, actually - because we're only accessing public data, any authenticated GitHub user can call the APIs needed. But yes, I agree needing to generate a GitHub token agrees the barrier to entry over not!

PR #115, opened 2023-06-16
Make cursor into hand when hovering over summary

illicitonion:
It's not super clear right now from styling these are clickable


SallyMcGrath:
You know we're not supposed to. But at the same time, for THIS audience, it's how VSCode works -- the twirldowns have hand pointers.  

But it's really not meant to be... https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

But marker is very limited - it only has a few available properties - chiefly it doesn't have transforms, so while it's easy to change color it's hard to give a secondary signal (colour can never be the only signal). OH I AM TORN.

Go on then...

SallyMcGrath (approved):
A bit reluctantly? But I do agree the clickiness is possibly missable

PR #118, opened 2023-06-16
Feature revision CodeMirror : add oneDark and merge with other code blocks

SallyMcGrath:
1. add oneDark theme
2. move over pre-existing surround to match through with other codeblocks

I don't think we need to do very much more than this


assets/styles/04-components/codemirror.scss
40thieves, line 1: I'm not sure this quite has the desired effect?

Screenshot 2023-06-25 at 21 36 18

See [here](https://deploy-preview-118--cyf-curriculum.netlify.app/guides/code-style-guide/#essential)

assets/styles/04-components/codemirror.scss
```diff
@@ -0,0 +1,14 @@
+.cm-scroller {
+  display: block;
```
40thieves, line 2: This appears to be overwritten by CM6 built in styles.

40thieves (approved):
The one dark theme LGTM, but I think the code block styles aren't quite right?

I'm fine with merging this now and we can follow up with fixes for the styles later (happy to write up a ticket if you'd like)

40thieves:
Also, apologies for the delay in reviewing!

SallyMcGrath:
I didn't try to make it identical, just sort of in the same block size with outline. I didn't set an overall theme - that's a bigger job really.

SallyMcGrath:
Screenshot 2023-06-27 at 21 31 05

40thieves:
Once again, sorry for being late to the party here but that looks good to me.

I wasn't sure if the intention was to match the other styles, but I'd also noticed that there were some styles that conflicted with those that CM6 applies and I wasn't sure what effect that would have. But this looks like it would be fine now 👍🏻

PR #119, opened 2023-06-16
markdownify is sometimes weird

SallyMcGrath:
https://github.com/gohugoio/hugo/issues/7372

If the text is just one line, markdownify doesn't give us paragraphs, or a way to tell if this is the case. In the PD blocks we quite often get this.

I am constraining phrasing blocks to a readable line length, so this choice has some text spooling out across the page

Therefore I have begrudgingly wrapped these risk areas in a div and constrained that.

But I don't like it

### DIFF
--- a/assets/styles/04-components/copy.scss
+++ b/assets/styles/04-components/copy.scss
@@ -3,7 +3,8 @@
 .c-copy {
   & p,
   & ol,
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Sure, I believe you

PR #126, opened 2023-06-18
invoke youtube partial from youtube shortcode

SallyMcGrath:
this is actually obviously the right thing to do but clearly I love to WET

this means we should write all our coolest funnest components and then invoke them from shortcodes or codefences to use them in markdown


illicitonion (approved):
Sure why not

Issue #131, opened 2023-06-19
Bug: Quirksmode


SallyMcGrath:
**Describe the bug**

I've deleted the doctype at some point and this site is running in quirksmode

**What is wrong with this content?**

Not much will change but it probably means the SVG will be weird when I change it back and I'll need to remake it

**How could this be fixed?**

1. Add back the doctype
2. Fix the small changes (most likely just the SVGs)

**Link to content**

**Additional context**

**Who else might need to know about this?**

PR #133, opened 2023-06-19
Remove Quirksmode and tidy up

SallyMcGrath:
Bugfix I guess?

Addresses #131

- added back the doctype
- fixed all the weird tiny layout bugs this caused
- remade the SVG which had gone crazy


illicitonion (approved):
I believe you! I'm sure anything weird will show up in a final proof :)

PR #134, opened 2023-06-20
PWA first iteration

SallyMcGrath:
I'll write this to run forward and cache groups of pages in sensible manner at some point but here's a basic functioning start.

test by looking in Application > Service Worker 
sw.js
activated and is running

check Application > Storage > Cache Storage > CYF caches exist

browse around and then turn off Network access, either by killing your connection or checking Service Workers > Offline

I can't test this on local as back forward cache is disabled with websockets


SallyMcGrath:
gonna merge so we can test

layouts/partials/head.html
```diff
@@ -25,4 +47,24 @@
     onload="this.media='all'" />
 
   <link rel="shortcut icon" href="{{ .Site.BaseURL }}favicon.ico" />
+  <meta name="theme-color" content="#e3ddeef2" />
+  <link rel="manifest" href="manifest.json" />
+  <script src="sw.js"></script>
```
40thieves, line 52: Sorry for being late to the party, but is this line necessary?

My understanding of service worker is that you load them via `register`, as you've done on line 56 below?

SallyMcGrath: I have a bigger issue with this worker actually, which is 

18sw.js:36 Uncaught (in promise) TypeError: Failed to execute 'clone' on 'Response': Response body is already used

I will do a revision one mo

PR #116, opened 2023-06-16
Add Fundamentals Iteration: Ship It content

illicitonion:
## What does this change?

Module: Fundamentals
Week(s): 2

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.MD)
- [x] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- [x] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
- [x] I have run my code to check it works
- [x] My changes follow our [Style Guide](https://curriculum.codeyourfuture.io/guides/code-style-guide)

## Description

Previously this block was a copy of the Requirements content, which is not what the trainees should be doing.

This is based on the content in
https://syllabus.codeyourfuture.io/fundamentals/week-2/session#step-by-step but slightly reworked.

Fixes #78

## Who needs to know about this?

@SallyMcGrath @gregdyke

PR #139, opened 2023-06-22
Fix previewing instructions in Fundamentals 2 Ship It

illicitonion:
## What does this change?

Module: Fundamentals
Week(s): 1

## Description

VSCode doesn't show previews by default (as was expected from https://syllabus.codeyourfuture.io/fundamentals/week-2/session#step-by-step
- though maybe we used to have trainees install extensions earlier or something? I couldn't find where, if so...)

So instead have the trainees explicitly install an extension.

Follow up from #116

### DIFF
--- a/content/fundamentals/blocks/ship-it-iteration/index.md
+++ b/content/fundamentals/blocks/ship-it-iteration/index.md
@@ -29,14 +29,16 @@ First you'll integrate the code you made in CYF Blocks into your website locally
 
 1. Open up VSCode (which you should already have installed).
 1. In VSCode, open up the folder where your website is saved (**File > Open Folder** and choose the unzipped folder).
-1. Now your IDE shows your code just like Codepen did. Take a look. You can also view the rendered view right in your browser. Take a look!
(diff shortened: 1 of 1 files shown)


textbook:
Extensions are covered over in https://docs.codeyourfuture.io/leaders/joining-the-course/getting-setup#5-visual-studio-code, we can add to https://github.com/CodeYourFuture/cyf-extension-pack

SallyMcGrath (approved):
Yep

illicitonion:
> Extensions are covered over in https://docs.codeyourfuture.io/leaders/joining-the-course/getting-setup#5-visual-studio-code, we can add to https://github.com/CodeYourFuture/cyf-extension-pack

If you could follow up on both that page and the extension pack that'd be really handy, thanks @textbook!

textbook:
It looks like `ms-vscode.live-server` is simlar to `ritwickdey.LiveServer`, which is already in our recommended list/extension pack. It's officially supported, but still in preview - maybe we should wait for GA before cutting over?

PR #144, opened 2023-06-22
Misc copy edits

illicitonion:
Misc copy edits


SallyMcGrath (approved):
no problems with this

PR #151, opened 2023-06-24
Set store in correct rendering context

### LINKED ISSUE #143: mermaid rendering is broken
On https://curriculum.codeyourfuture.io/fundamentals/prep/ the mermaid isn't rendering:

Screenshot 2023-06-22 at 22 00 37

### LINKED ISSUE #145: "Unplugged Version" tab is broken
On https://curriculum.codeyourfuture.io/fundamentals/sprints/1/day-plan/

Screenshot 2023-06-22 at 23 58 16

Clicking the "Unplugged Version" tab doesn't appear to do anything

SallyMcGrath:
We need to set the .Store on the top-level page loaded in Step 2, and we do that with the page function.

https://gohugo.io/variables/page/#the-global-page-function
https://discourse.gohugo.io/t/page-store-is-sometimes-empty-cannot-load-mermaid-script-reliably/44950/9

fix thanks to jmooring!

fixes #143 and fixes #145


illicitonion (dismissed):

illicitonion:
The "Unplugged Version" tab on https://deploy-preview-151--cyf-curriculum.netlify.app/fundamentals/sprints/1/day-plan/ is still broken

SallyMcGrath:
> The "Unplugged Version" tab on https://deploy-preview-151--cyf-curriculum.netlify.app/fundamentals/sprints/1/day-plan/ is still broken

Then I think I need to store in both contexts to be safe

SallyMcGrath:
This is very hard to debug as it's completely different on this preview and on my local.

SallyMcGrath:
I've torn out the Hugo approach and put in a loader like it's 2015.

SallyMcGrath:
So, this does now work, but I've set to draft as I can't find any evidence it's better or more performant, really, than just loading it as we are already.

40thieves:
If I understand it correctly, it will introduce a small waterfall where any individual components would have to wait for the `component.js` script to complete, then the "sub" scripts for that component.

But since it's a pretty small waterfall, and it seems there isn't a better alternative, I don't think we'd be paying a huge penalty. Perhaps worth running a comparison with Lighthouse just to be on the safe side?

SallyMcGrath:
> If I understand it correctly, it will introduce a small waterfall where any individual components would have to wait for the `component.js` script to complete, then the "sub" scripts for that component.
> 
> But since it's a pretty small waterfall, and it seems there isn't a better alternative, I don't think we'd be paying a huge penalty. Perhaps worth running a comparison with Lighthouse just to be on the safe side?

Yep, I did test it and could find no performance improvement. So it doesn't seem worth it.

PR #183, opened 2023-07-22
Feature : nestable Youtube shortcode, in new tabs

SallyMcGrath:
## What does this change?

This is a bugfix but also a feature change tbh as it takes out a bunch of (unused) functionality 

I know we've all pointed it out, I can't find the issue for it!

## Checklist

-[x] My changes follow our [Style Guide](https://curriculum.codeyourfuture.io/guides/code-style-guide)

## Description

I have made these changes:

In the youtube shortcode partial, 

which is a pass-through that calls the youtube block, I have also passed along the context that this is a shortcode by adding it to the blockData. I didn't add it to the page scratch as I thought it might get messy with multiple versions of this block on the page, some shortcode some not.

You can test this on fundamentals/prep/ which has a video in a tab up top and then an inline video down the page. The inline video has a (visually hidden) heading and the tab video does not.

I then stripped out all the page bundle logic from the tabs. We aren't using it, we don't really understand it, so let's take it out until we want it.

### DIFF
--- a/layouts/partials/block/data.html
+++ b/layouts/partials/block/data.html
@@ -10,6 +10,8 @@
   3. block name
   4. source of truth (sot)
   5. headers and auth for the api call
+  6. showTime, a boolean to show the time in the block
+  7. isShortcode, a boolean to indicate if the block is being called from a shortcode
 */}}
 {{ $name := .name | default "warning: name is empty" }}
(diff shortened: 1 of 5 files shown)


SallyMcGrath:
I really don't know why that content page is in there. It's really a mystery to me.

illicitonion (approved):
Simpler is better!

PR #198, opened 2023-07-26
Bug: weird escaping markdownifying inner nested shortcodes

SallyMcGrath:
1. Make tabs fully work with youtube shortcode
- 1. you have to declare the shortcode as safeHTML if you're nesting this deep but only sometimes so I've set it for all cases as this is always html
- 2. You have to actually set the boolean isShortcode to the scratch map instead of just somehow willing it with your mind

2. Store this one weird tip by [showing content change](https://github.com/CodeYourFuture/curriculum/pull/198/files#diff-e4a97e4548b6599d18a607fdde1c1fdb92273bab764dc55a2ad6ff3176a05b03)

If you use % markers in markdown in nested tabs your html will always be escaped, safeHTML or nay, so don't use those unless you want that, lesson learned

### DIFF
--- a/content/fundamentals/prep/index.md
+++ b/content/fundamentals/prep/index.md
@@ -26,7 +26,7 @@ name="Backlog"
 +++
 
 {{<tabs name="prep">}}
-{{% tab name="Getting Started" %}}
+{{< tab name="Getting Started" >}}
 
 ## Getting Started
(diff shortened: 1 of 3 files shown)


Dedekind561 (approved):
LGTM

PR #213, opened 2023-08-02
Content: HTML CSS Week 2

SallyMcGrath:
## What does this change?

Module: HTML CSS
Week(s): 2

## Checklist

## Description

add existing w3 playlist moved to w2 (forms moved to w3)
add 10 questions about css
asking questions workshop


Dedekind561:
@SallyMcGrath Did you push any more changes or is the PR ready for review again?

Dedekind561:
Actually I've just seen @40thieves you're assigned to this now - are you going to add a few more changes?

SallyMcGrath:
I made the small changes we discussed in SeeSaw immediately - this is ready

content/html-css/sprints/2/day-plan/index.md
```diff
@@ -7,8 +7,23 @@ weight = 3
...
 backlog_filter= 'Week 2'
 [[blocks]]
+name="energiser"
+[[blocks]]
 name="Active listening"
 src="https://cyf-pd.netlify.app/blocks/active-listening/readme/"
+[[blocks]]
+name="morning-break"
+[[blocks]]
+name="Asking questions"
```
Dedekind561, line 17: The day plan view is saying the mermaid diagram contains a Syntax Error for the Asking Questions workshop: however it renders properly in the GIthub UI 🤔.
Happy for this PR to go through if we raise an issue to fix this in another PR - unless you can spot a quick fix now.

Dedekind561 (approved):

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+You learn them by using them. You look it up here: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference. And the other thing you can do is look at the autocomplete in Devtools.
+
+</details>
+
+<details>
+<summary>Why is CSS a 'cascading' style sheet? What does cascade mean?</summary>
+
+<iframe src="https://giphy.com/embed/zFSGphnnT9opa" width="480" height="480" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/zFSGphnnT9opa">via GIPHY</a></p>
+
+A cascade is a stream or sequence of a things where each thing affects the next. You might imagine pouring champagne into a tower of glasses. The champagne flows into the containers; it cascades. In CSS we’re actually pouring from many bottles into one tower.
```
40thieves, line 52: Nitpick, I think this reads a bit better?

```suggestion
A cascade is a stream or sequence of a things where each thing affects the next. You might imagine pouring champagne into a tower of glasses. The champagne flows into the top container; it then cascades into the next container; etc. In CSS we’re actually pouring from many bottles into one tower.
```

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+So in CSS 'the cascade' means the order or sequence of the rules as they flow into the DOM tree according to CSS precedence.
+
+</details>
+
+<details>
+<summary>What does precedence mean?</summary>
+
+It means hierarchy. One hierarchy in CSS is inheritance: children inherit from parents. If a parent node has red text, all its children will inherit that red text color. But when another rule comes after, and overrules, it takes precedence. Everything in CSS is determined by order.
+
+Order is so important in CSS because CSS is a declarative programming language that programs the layout of boxes.
```
40thieves, line 63: Really nitpick-y here again, sorry:

```suggestion
Order is so important in CSS because CSS is a declarative programming language that programs the style of boxes.
```

I'd be fine with "layout and style of boxes" too.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+A cascade is a stream or sequence of a things where each thing affects the next. You might imagine pouring champagne into a tower of glasses. The champagne flows into the containers; it cascades. In CSS we’re actually pouring from many bottles into one tower.
+
+So in CSS 'the cascade' means the order or sequence of the rules as they flow into the DOM tree according to CSS precedence.
+
+</details>
+
+<details>
+<summary>What does precedence mean?</summary>
+
+It means hierarchy. One hierarchy in CSS is inheritance: children inherit from parents. If a parent node has red text, all its children will inherit that red text color. But when another rule comes after, and overrules, it takes precedence. Everything in CSS is determined by order.
```
40thieves, line 61: I think the "one hierarchy" is tripping me up since it implies that there are other hierarchies (which aren't explicitly called out).

How about something like this?

```suggestion
It means hierarchy.

One example of a hierarchy in CSS is inheritance: children inherit from parents. If a parent node has red text, all its children will inherit that red text color. Another example of a hierarchy is when another rule comes after, and overrules, it takes precedence. 

Everything in CSS is determined by order.
```

40thieves: I think if this wasn't recap material I would suggest an example (or at least a diagram) here.

But on reflection, trainees should be familiar with the concrete example and I think we're more interested in the abstract definitions in this exercise.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+It means, broadly speaking, that the entire code is processed and then applied, from top to bottom in one go. You don’t write for loops or if/else statements; there are no functions as such; you just _declare_ the outcome.
+
+In CSS you do the 'what', and the browser does the 'how'.
+
+</details>
+
+<details>
+<summary>What is specificity?</summary>
+
+Specificity is a [weighting system for CSS rules](https://specifishity.com/). The more specific your selector, the more power it has. For example: IDs can only ever mean _one_ node. IDs are unique, so they are _more specific_ than classes, which could mean many nodes.
```
40thieves, line 81:
```suggestion
Specificity is a [weighting system for CSS rules](https://specifishity.com/).

The more specific your selector, the more power it has. For example: an `id` attribute can only ever be attached to _one_ node. This means IDs are unique, so they are _more specific_ than classes, which could mean many nodes.
```

I also wonder if it would be worth swapping out "node" for "element" here? Above we use "element", and my impression is that is the more broadly understood term here.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+Classes added to html element are _more specific_ than just a plain element selector, as a class of `p` is _more specific_ than just all `p`. If you get two elements with the same specificity then the one that comes last in the order wins.
```
40thieves, line 83: I'm struggling to understand this point?

I think it's possibly the phrase "class of `p`": initially I thought you were using an example something like this: `<p class="p">...</p>`, which is quite confusing.

But re-reading it I think you're using "class" in the abstract sense here: all `p` in a (potential) class? Assuming that's true, the word "class" is quite overloaded here: it refers to the literal keyword and the abstract concept.

What about rewording to something like:

```suggestion
Classes added to html elements are _more specific_ than just a plain element selector, as a `p` with a class is _more specific_ than just all possible `p`s. If you get two elements with the same specificity then the one that comes last in the order wins.
```

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+The browser sorts all your rulesets, or declarations, into a defined order according to the power, or weight, of the declaration and applies the result, like a game of cards where higher value cards beat lower value cards.
```
40thieves, line 85: Given our syntax definitions above, I don't think "declarations" are sorted - only rulesets?

```suggestion
The browser sorts all your rulesets into a defined order according to the power, or weight, of the ruleset and applies the result, like a game of cards where higher value cards beat lower value cards.
```

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+Stronger rules _override_ weaker rules. Everything is sorted and sorted until eventually a final value wins out and the view is rendered. The browser sorts like this:
+
+1. Match the media
+1. Sort by origin
```
40thieves, line 90: Assuming that these refer to stylesheet loading, I wonder if they're just distractions and should be removed?

99% of the time, our trainees wouldn't come across these.

EDIT: see my comment below which expands on this.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+1. Sort by specificity
+1. Sort by document **order**
+
+</details>
+
+<details><summary>What do we mean by order?</summary>
+
+Order means 'what comes first?'.
+
+Another way to think of ‘earlier and later’ is ‘farther and nearer’. The _closer_ the declaration is to the node in the DOM, the ‘later’ it is in the CSS. So an inline style overrides a style set in a style block, which overrides a stylesheet linked in the head, which overrides any stylesheet linked higher up in the head, which all ultimately overrides the user agent style that comes in the browser.
```
40thieves, line 100: This is really interesting... I don't think about order in this way, instead I think about order within the stylesheet.

I think that actually might be more relevant to our trainees too? My expectation is that in general our trainees would be operating on a single stylesheet so I would suggest adding something about order within the stylesheet.

I'd also suggest that details of multiple stylesheet loading are not very relevant and so could be removed?

40thieves: Also have we introduced the concept of a user agent stylesheet before?

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+</details>
+
+<details>
+<summary>What does specify mean?</summary>
+
+Identify clearly. Name explicitly.
+
+</details>
```
40thieves, line 109: I feel like this is lacking context? It's the only time we use the word "specify" on the page.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+<details>
+<summary>What does render mean in context of a browser?</summary>
+
+Render means to hand over, actually - to give back. In the browser, it means to show the result of all these computations we have talked about. It turns the code into pixels and paints them on your screen. So it’s the end result - the final face.
```
40thieves, line 114: Not sure I know what "the final face" means?

40thieves (approved):
I'm going to approve as we're officially past publish deadline here, but I've left quite a few suggestions.

Would you like me to turn those into a follow up PR? (Unfortunately unlikely to get to that for a few days at least)

40thieves:
For reference, I didn't review the asking questions workshop. I'll try to look at that separately.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+[objectives]
+    1='Define all key terms of CSS'
++++
+
+{{<note type="question" title="10 Things About CSS">}}
+
+<details>
+<summary>What is CSS?</summary>
+
+Cascading Style Sheets are a language of style - a visualisation of the DOM - using selectors, properties, and values in rulesets.
```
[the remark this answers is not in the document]
SallyMcGrath: It's calling back to What is HTML from last week.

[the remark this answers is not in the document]
40thieves: Ah of course, I totally forgot this week 2 🤦‍♂️

content/html-css/blocks/what-is-css/index.md
SallyMcGrath: I definitely want declaration in there. Can you give another revision?

content/html-css/blocks/what-is-css/index.md
SallyMcGrath: Yes we do - it's in Devtools workshop last week and also in the video they watch before this section.

content/html-css/blocks/what-is-css/index.md
SallyMcGrath: It's a common question that comes up every time we do this module.

content/html-css/blocks/what-is-css/index.md
```diff
@@ -0,0 +1,118 @@
...
+headless = true
+time = 60
+facilitation = false
+vocabulary=['property', 'value', 'cascade', 'specificity', 'declarative', 'render', 'block', 'inline']
+emoji= '🧩'
+[objectives]
+    1='Define all key terms of CSS'
++++
+
+{{<note type="question" title="10 Things About CSS">}}
```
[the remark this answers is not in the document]
SallyMcGrath: If you read the page in the rendered view, you will get the context. It does have context and clear instructions in what to do, but it's just not in this little chunk here. 

It's the same exercise as the previous week, as well.

content/html-css/blocks/what-is-css/index.md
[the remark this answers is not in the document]
SallyMcGrath: Oh yes so it's useful to say, this is w2 of sdc, which means these trainees have 

- Built 5 full websites deployed to Netlify over 4 weeks
- Written a personal home page with tests
- Developed and deployed a new site over 3 weeks
- Begun Git 
- Built 2 more pages in the last week

This is **week 10** of their engagement with HTML and CSS

content/html-css/blocks/what-is-css/index.md
SallyMcGrath: Hmm, good point. Let's merge this and then do another revision

PR #221, opened 2023-08-03
fix readme timestamp on day-plan schedule

Dedekind561:
## What does this change?

Currently the time render properly for workshops on the day plan.
Just tweaked the data passed to the time partial so it shows up properly.

### Currently, the day-plan looks like this in production:
image
Note: no time for the git workshop section.

### Now checkout the same page on the deploy preview:
https://deploy-preview-221--cyf-curriculum.netlify.app/induction/sprints/1/day-plan/#git-workshop
image
It should have the correct time and it should have a duration of 2 hours.

## Checklist

### DIFF
--- a/layouts/partials/block/local.html
+++ b/layouts/partials/block/local.html
@@ -8,7 +8,7 @@
     <h2 class="c-block__title e-heading__2" id="{{ $blockData.name |urlize }}">
       {{ $block.Title }}
     </h2>
-    {{ partial "time.html" (dict "blockData" $blockData "time" $block.Params.time) }}
+    {{ partial "time.html" (merge $blockData (dict "time" $block.Params.time)) }}
   </header>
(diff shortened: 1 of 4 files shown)


SallyMcGrath (changes requested):
I think that instead of making this change here, the change should be in the time partial, which is currently accessing a .time argument but should access blockData.time from /block/data.html

Dedekind561:
Cool, I've updated the time partial to read `time` property off the map that's passed to the time partial. I also merge the blockData and time attributes into a single map and then pass that to the partials in the block components.

SallyMcGrath:
I will move my revision to a new PR and stop blocking this on

SallyMcGrath (approved):
approved with the caveats documented - this is actually broken on live tho so this will do for now!!

PR #241, opened 2023-08-13
Copy Edits: substantial JS1-1

SallyMcGrath:
## What does this change?

Module: JS1
Week(s): 1

Hey this won't work as it's dependent on tooltips and maybe also some work on images that I might do... in my copious spare time.

We should talk about my changes here, as they are  fairly large. I moved them to a new branch so it is easier to side by side.

My edits including code make this text about 2.9k words. 

The version you submitted was about 3.5k words, so it's not a massive haircut. I moved some images into code and created a tooltip component to put some definitions in. 

I broke the whole thing into 9 blocks and composed them, which I think will make editing easier in the future.


Dedekind561:
@SallyMcGrath - your edits are looking great, thankyou! The tooltips are very cool 😎. 
Is there stuff you want to discuss further on this PR - I'm happy for this to get merged and then I can do some mop up 🧹  PRs afterwards.

SallyMcGrath:
If you're happy with the content changes I made, then maybe we don't need to discuss. 

I think I have broken images inside tabs though, or discovered they are broken, rather, by using them loads here. I think I'd rather fix that than remove images in tabs. What do you think?

Dedekind561:
Yeah no problem - if you want to fix that I can hold off for now. I can store my other tweaks somewhere else and update some other branches.

SallyMcGrath:
I don't think my fix should block this, as the bug wasn't introduced in this PR, just discovered as a side effect

Dedekind561:
LGTM - I'll mop up in other PRs

Dedekind561 (approved):

PR #244, opened 2023-08-16
Rewrite js2 week 1 prep content

Dedekind561:
## What does this change?

Module: JS2
Week(s): 1

## Checklist


carl0FF (commented):
@Dedekind561, you made things flow amazingly on this one! Great work!

content/js2/blocks/reference/index.md
```diff
@@ -0,0 +1,91 @@
...
+We can also check these variables share the same reference.
+
+```js
+const list = [10, 20, 30];
+const copy = list;
+
+console.log(list === copy); // evaluates to true
+```
+
+If we're comparing 2 variables with `===`, then it will check to see if they have the **same reference**.
```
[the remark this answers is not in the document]
Dedekind561: Cool, sounds good

content/js2/blocks/side-effects/index.md
```diff
@@ -0,0 +1,34 @@
...
+```js
+test("doesn't modify the input", () => {
+  const list = [1, 2, 3];
+  calculateMedian(list);
+
+  expect(list).toEqual([1, 2, 3]);
+});
+```
+
+In this test, we don't check the **return value** of `calculateMedian`. We assert that the input has the same contents as the original input. We can use the `toEqual` matcher to check the contents of the original input.
```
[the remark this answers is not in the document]
Dedekind561: The "Fail Fast Prep" section is the PD section of the prep material.
But yeah there is an abrupt end to the tech-ed section. There is going to be a "wrap up" block at the end of the tech-ed prep section, which will end the section a bit more smoothly

content/js2/blocks/mean/index.md
```diff
@@ -0,0 +1,37 @@
...
+_When_ we call `calculateMean` with the array of numbers
+_Then_ we get the mean
+
+Let's create a test to check its functionality:
+
+```js
+test("calculates the mean of a list of numbers", () => {
+  const list = [3, 50, 7];
+  const currentOutput = calculateMean([3, 50, 7]);
+  const targetOutput = 20;
```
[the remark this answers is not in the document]
Dedekind561: I was concerned about this - I'm going to leave as it is for now ( because I'd have to go through and update it in other places too! ). However, we can review this again later on if people find the naming is unclear here.

content/js2/blocks/assembly/index.md
```diff
@@ -0,0 +1,49 @@
...
+const salaries = [10, 20, 30, 40, 60, 80, 80];
+const median = calculateMedian(salaries);
+const mean = calculateMean(salaries);
+
+console.log(`The median salary is ${median}`);
+console.log(`The mean salary is ${mean}`);
+```
+
+{{<note type="activity" title="Run it">}}
+Try running the code above on your local machine. Check the console output. What do you see?
```
illicitonion, line 23: I think we should either have the trainees _predict and write down_ an expectation, or we should tell them what our expectation is. Otherwise just telling them "What do you see?" doesn't necessarily prompt that something is wrong. Let's keep explicitly training "predict + observe" as a behaviour.

content/js2/blocks/iteration/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+{{<tabs name="Playing computer">}}
+
+{{<tab name="🕹️ Play computer">}}
+
+**Play computer** with the code below and explore what happens the `for...of` loop is executed.
+
+Pay close attention to what happens when we enter the `for...of` loop
+
+<iframe title="for..of loop" width="800" height="500" frameborder="0" src="https://pythontutor.com/iframe-embed.html#code=function%20calculateMean%28list%29%20%7B%0A%20%20let%20total%20%3D%200%3B%0A%20%20for%20%28const%20item%20of%20list%29%20%7B%0A%20%20%20%20total%20%2B%3D%20item%3B%0A%20%20%7D%0A%7D%0A%0AcalculateMean%28%5B10,20,30,40,50%5D%29%3B&codeDivHeight=400&codeDivWidth=350&cumulative=false&curInstr=0&heapPrimitives=nevernest&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false"> </iframe>
```
illicitonion, line 35: The frames here suggest a distinct `item` variable is declared in the function scope - I don't _think_ this is the case? Do you know what the deal is here?

content/js2/blocks/median/index.md
```diff
@@ -0,0 +1,70 @@
...
+    A[Step 1: Find the middle index of the array] --> B[Step 2: Get the middle item]
+    B --> C[Step 3: return the middle item]
+```
+
+In code we can we can use [`splice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) to to get the middle item.
+
+```js
+function calculateMedian(list) {
+  const middleIndex = Math.floor(list.length / 2);
+  const median = list.splice(middleIndex, 1)[0];
```
illicitonion, line 50: I think jumping straight to a buggy side-effecting implementation using `splice` as our first example is probably a bit confusing and training a bad habit...

What do you think about swapping the order of the `mean` and `median` blocks, so that we're teaching expected good behaviour before introducing bugs?

content/js2/blocks/reference/index.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+In the example above, `salaries` is assigned a reference on the first line.
+Explain why `calculateMedian` and `calculateMean` both get access to the same array.
+
+{{</tab>}}
+
+{{</tabs>}}
+
+### Shared reference
```
illicitonion, line 80: I think we should explicitly introduce contrasts here with by-value types, both for the above (`+=`ing to an int or string in a function doesn't affect it, because values not references) and below (`===`ing different ints or strings _does_ evaluate to `true`).

content/js2/blocks/reference/index.md
```diff
@@ -0,0 +1,91 @@
...
+We can also check these variables share the same reference.
+
+```js
+const list = [10, 20, 30];
+const copy = list;
+
+console.log(list === copy); // evaluates to true
+```
+
+If we're comparing 2 array variables with `===`, then it will evaluate to `true` only if they have the **same reference**. `===` is comparing the references to the arrays, not the arrays themselves.
```
illicitonion, line 91: Agains, we should add contrasts here - that strings and numbers _have a different definition of equality_.

I think we should also call out explicitly that two arrays that happen to have the same contents don't compare equal. In particular, calling out that `console.log(list === [10, 20, 30])` logs `false` is probably worthwhile.

content/js2/blocks/side-effects/index.md
```diff
@@ -0,0 +1,29 @@
...
```
illicitonion, line 29: Is there a missing exercise here to fix the `calculateMedian` implementation?

If not, should we give a sample implementation which is simpler and less buggy?

Either way, I think we should explicitly mention that the implementation we supplied was intentionally buggy as a learning experience, and that people shouldn't use `splice` as a way to access array elements by index.

illicitonion (commented):
This looks great, thanks!

I'm not sure about the tiny blocks, as I suspect it'll make it a bit harder to edit/reorder stuff in the future, but I don't have strong opinions.

content/js2/blocks/iteration/index.md
Dedekind561: That looks like a bug to me :(

Dedekind561: Maybe worth switching to the VSCode debugger?

content/js2/blocks/median/index.md
Dedekind561: Yeah that makes sense - definitely think we can swap those blocks.

content/js2/blocks/reference/index.md
Dedekind561: Hey there, could you provide an example of `+=` to an int or string not being affected? I'm not sure what you mean here. Do you mean trying to reassign a string/int value that's passed as an argument to a function?

content/js2/blocks/iteration/index.md
illicitonion: Yeah, I followed up, it's a known issue in [the Python Tutor docs](https://docs.google.com/document/d/13_Bc-l2FKMgwPx4dZb0sv7eMfYMHhRVgBRShha8kgbU/edit):
> for-loop variables show up duplicated in two nested blocks due to how the Node.js JavaScript debugger emits its values ([example](https://pythontutor.com/visualize.html#code=//%20%5Bofficial%20Python%20Tutor%20example%5D%0A%0A//%20notice%20how%20there%20are%202%20copies%20of%20the%20i%20and%20j%20loop%20variables%20in%20visualizer%3A%0Afor%20%28let%20i%20%3D%200%3B%20i%20%3C%2010%3B%20i%2B%2B%29%20%7B%0A%20%20for%20%28let%20j%20%3D%200%3B%20j%20%3C%20i%3B%20j%2B%2B%29%20%7B%0A%20%20%20%20console.log%28i,%20j%29%3B%0A%20%20%7D%0A%7D&cumulative=false&heapPrimitives=nevernest&mode=edit&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false))

If you think there's a convenient way to get folks to use a debugger here, that sounds great, otherwise I guess we can just drop the visualisation (I don't think it's worth including the buggy visualisation and adding a caveat, that'd probably be more confusing than it is useful at this stage).

content/js2/blocks/reference/index.md
illicitonion: Contrasting examples of ~identical code with different types:

```javascript
function addSToArray(arr) {
  arr.push("s");
}
const lettersInAnArray = ["a"];
addSToArray(lettersInAnArray);
// Because `lettersInAnArray` was passed by reference, `addSToArray`'s modification is visible here, because the same underlying storage was modified.
console.assert(lettersInAnArray.length === 2);
console.assert(lettersInAnArray[1] === "s");

function addSToString(str) {
  str += "s"
}
const string = "a";
addSToString(string);
// Because `string` was passed by value, `addSToString`'s modification is not visible here, because a copy was made just for the function before the value was modified.
console.assert(string.length === 1);
console.assert(string === "a");
```

These two functions look the same - they both take an argument, `.push` and `+=` are logically equivalent operations. But they behave differently because of passing by reference vs value.

This is spooky and in my experience hard to get your head around, so the more examples we can give here, and the more we can specifically call out "in JS: Arrays and objects are reference types, everything else is a value type", the more clear we can hopefully make this.

content/js2/blocks/side-effects/index.md
```diff
@@ -0,0 +1,55 @@
...
+function calculateMedian(list) {
+  const middleIndex = Math.floor(list.length / 2);
+  const median = list.splice(middleIndex, 1)[0];
+
+  return median;
+}
+```
+
+We've established that we shouldn't use `splice` to retrieve the median from the input array.
+Fix the implementation of `calculateMedian` above so it no longer calls `splice` and mutates the input.
```
illicitonion, line 47:
```suggestion
Fix the implementation of `calculateMedian` above so it no longer calls `splice` (which mutates the input), and instead gives the right answer _without_ mutating the input.
```

content/js2/blocks/iteration/index.md
Dedekind561: Thanks for following this up - yeah the buggy visualisation will just end up being confusing. I'm going to get rid of it 🧹

Dedekind561:
@illicitonion I've just pushed two final updates as per your comments. I've linked to a video I've just made on using the debugger to visualise the `for...of` loop too. I've also added a section on value vs reference. I've used two different implementations of a function `pluralise` ( same as yours ) side by side in tabs so trainees can compare the two different behaviours for themselves.
Let me know once you've checked and made final suggestions - then I'll get this merged :)

content/js2/blocks/reference/index.md
```diff
@@ -0,0 +1,130 @@
...
+<iframe title="pluralise-string" width="1100" height="500" frameborder="0" src="https://pythontutor.com/iframe-embed.html#code=function%20pluralise%28arr%29%20%7B%0A%20%20arr.push%28%22s%22%29%3B%0A%7D%0Aconst%20lettersInAnArray%20%3D%20%5B%22c%22,%22a%22,%22t%22%5D%3B%0Apluralise%28lettersInAnArray%29%3B%0A%0Aconsole.assert%28lettersInAnArray.length%20%3D%3D%3D%204%29%3B%0Aconsole.assert%28lettersInAnArray%5B3%5D%20%3D%3D%3D%20%22s%22%29%3B&codeDivHeight=400&codeDivWidth=600&cumulative=false&curInstr=0&heapPrimitives=nevernest&origin=opt-frontend.js&py=js&rawInputLstJSON=%5B%5D&textReferences=false"> </iframe>
+{{</tab>}}
+
+{{<tab name="pluralise: string">}}
+
+Here `pluralise` is **passed a string by value**.
+
+This means a copy of `string`'s value is passed to `pluralise` in the second tab. `pluralise`'s reassignment is not visible here, because a copy was made just for the function before the value was modified.
+
+Step through the code to observe this behaviour:
```
illicitonion, line 125: The code in this one is buggy - the assertions are:

```js
console.assert(string.length === 1);
console.assert(string === "a");
```

where they should be:
```js
console.assert(string.length === 3);
console.assert(string === "cat");
```

illicitonion (approved):
Looks great, thanks so much! Great video!

PR #267, opened 2023-09-08
Feature: PR Block

SallyMcGrath:
## What does this change?

Module: all
Week(s): all

## Description

This is NOT a shortcode but a block. It's really hard to make this a shortcode because the main value of making it a shortcode is to dynamically route to whatever module the block is showing up on. The compile order of Hugo means this shortcode doesn't know where it's showing, ultimately, so while I'm sure it's not impossible to find out, it's much much harder.

I have written this as a block instead.

iIf you do want to make it a shortcode, I'll include the basic code in the comments and you can wrestle with the context passing problem. :) (It's easy to make a shortcode where you pass the path directly, but also maybe pointless?)

The goal of this block is to prompt CYF community members to do code reviews during class days. I have set the time to zero so it can be added in to any day plan (eg after a study group block) without mucking up the times. 

If you want to put the time partial in, I've commented with how to do this in the partial. Crack on!

## Rendered Pages

https://deploy-preview-267--cyf-curriculum.netlify.app/html-css/sprints/1/day-plan/#code-review


SallyMcGrath:
Screenshot 2023-09-08 at 16 50 53

SallyMcGrath:
```shortcode
{{ $src := .Inner | default "https://github.com/CodeYourFuture/Module-HTML-CSS/pulls" }}
{{ partial "block/data.html" (dict "Scratch" .Page.Scratch "src" $src  "name" "" "page" .) }}
{{ $blockData := .Page.Scratch.Get "blockData" }}
{{ .Page.Scratch.SetInMap "blockData" "isShortcode" true }}
{{ partial "block/pullreq.html" . }}
```

illicitonion (commented):
This is cool!

A couple of questions:
1. Is there a nice way to make this filter / group by city? Maybe we could tab by city? Or add a "choose a city" drop-down? As a volunteer attending class in London, ideally I want to be filtering to the London PRs. If we assume people are using the correct PR titles, this feels possible?
2. Is there a cute way to make this dynamic? Ideally this is a piece of content we'd want to fresh load on page-load, rather than be stale based on when the site happened to publish... I'm not sure there's a nice way to do that, though?

SallyMcGrath:
> This is cool!
> 
> A couple of questions:
> 
> 1. Is there a nice way to make this filter / group by city? Maybe we could tab by city? Or add a "choose a city" drop-down? As a volunteer attending class in London, ideally I want to be filtering to the London PRs. If we assume people are using the correct PR titles, this feels possible?

If these were labelled, it would be super easy to group by label as a static component or a dynamic one. It would probably be better to write a labeller as a github action than parse and group in Hugo based on title strings. (It's not impossible to do this, but it would be a globby chunk of logic in a template view and ideally we'd limit that much business logic in templates. )

In general, as different cohorts are going through at different times, it will do this implicitly, as you will be looking at the html day plan when your cohort is doing html, and the js one when they're doing that, and it's picking the top 5. I would like a labeller though, because I'd like the action to reject PRs that aren't titled correctly. If you write that action I will revise this as a web component. (And link your example in [this Project Briefing](https://docs.google.com/document/d/1-_uqmwXzOhLtdO0pJfIutJEfvAuKH5EwdnC2kNrxZTc/edit?usp=sharing))?

> 2. Is there a cute way to make this dynamic? Ideally this is a piece of content we'd want to fresh load on page-load, rather than be stale based on when the site happened to publish... I'm not sure there's a nice way to do that, though?

2. Yes we can make it dynamic by writing it as a web component and doing the fetch inside the component instead of at build.  This makes the page heavier but not too bad. 

Let's merge this as MVP and write a new ticket with the new features?

SallyMcGrath:
Loads of stuff backing up here, so I'm merging and will open a new ticket

PR #270, opened 2023-09-08
Content: HTML module project

SallyMcGrath:
## What does this change?

Module: HTML-CSS
Week(s): 1-4

## Checklist

## Description

1. I connected the module project issues as block calls to the product pages
2. Probably for JS1 we will want a repo with plan/build/test/ship as folders with readmes instead but this seems ok for this module
3. Doing this exposed that I failed to update the block call in single.html, the layout all singles fall back to if they don't have a named layout, so I fixed that path ( I have a zillion PRs open so I'm not splitting it out as I caused confusion last time, though I do think content and features should be separate ideally...)

### DIFF
--- a/content/html-css/product/_index.md
+++ b/content/html-css/product/_index.md
@@ -1,9 +1,7 @@
 +++
 title = 'Product'
-description = 'Product description'
+description = 'Over one month, produce a high quality UI that meets the brief'
 layout = 'product'
 emoji= '🎁'
 menu_level = ['module']
 +++
-
-
--- a/content/html-css/product/build/index.md
+++ b/content/html-css/product/build/index.md
@@ -1,10 +1,11 @@
 +++
 title = 'build'
-description = 'build description'
+description = 'Iterating on your product, adding new features and expanding your codebase'
 layout = 'build'
 menu_level = ['product']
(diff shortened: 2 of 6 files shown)


SallyMcGrath:
Just so much stuff backing up without response I am merging without review.

If you want changes, open a PR

illicitonion:
This looks great!

I'm slightly confused on the "ship" page for two reasons:
1. It's not super obvious how it's connected to the project - is this just a reminder of the guide?
2. Ideally they'd be deploying from week 1, and the test/review portion would include deploy previews

What do you think about folding the "ship" into the "plan" section, by making it a link in https://github.com/CodeYourFuture/Module-HTML-CSS/issues/4 and removing the "ship" link?

I'm happy to make the changes if they sound good.

SallyMcGrath:
Let's think a bit more about how we want to format this project view. I don't think it is comfortably housing us at the moment.

We do want it to be really clear they must ship their work. It must be deployed. I agree we want them deploying from week 1. 

We don't have to have this format - we are not committed to it yet (the rest of the module structure, yes, but this, not yet). What could work better? Requirements Iteration Delivery?

PR #273, opened 2023-09-10
Copy edits

illicitonion:


content/js2/blocks/multiple-params/index.md
```diff
@@ -74,3 +78,28 @@ function parseQueryString(queryString) {
...
+    queryParams[key] = value;
+  }
+
+  return queryParams;
+}
+```
+
+When we're solving problems involving several values, often we need slightly differently handling for the cases when there are 0, 1, or more than 1 values. In our example here, we need to treat 0 values specially (if the query string is empty, we return early), but we can handle 1 and more than 1 the same way.
+
+When you're breaking down problems, think to yourself: What are special cases we may need to handle differently?
```
illicitonion, line 105: @Dedekind561

Can we also talk about edge-cases here? Three important ones to identify are missing =s (`foo&bar=baz`), repeated keys (`foo=bar&foo=baz`), and empty segments (`?&foo=bar`). I think we should probably have a section on identifying the edge-cases, then set an exercise to work out how to handle them, and implement that?

content/js2/blocks/access/index.md
```diff
@@ -77,3 +90,54 @@ Predict and explain what the console output be when we run the code above runs.
...
+  age: 33,
+};
+console.log(profileData.cityOfResidence);
+
+profileData.cityOfResidence = "Manchester";
+
+console.log(profileData.cityOfResidence);
+```
+
+Predict and explain what the console output be when we run the code above runs.
```
Dedekind561, line 113:
```suggestion
Predict and explain what the console output will be when we run the code above runs.
```

content/js2/blocks/access/index.md
```diff
@@ -42,6 +42,19 @@ const profileData = {
 console.log(profileData.firstName); // logs "Francesco"
 ```
 
+Objects also allow looking up property values using square brackets, similar to arrays. Instead of an index, we use a string of the key of the property inside the square brackets:
```
Dedekind561, line 45:
```suggestion
Objects also allow looking up property values using square brackets, similar to arrays. Instead of an index, we use a string of the key inside the square brackets:
```

content/js2/blocks/multiple-params/index.md
```diff
@@ -28,8 +28,12 @@ test("given a query string with multiple key-value pairs, returns them in object
 
 We've already worked out how to update the query params object given a **single key-value pair** in the query string.
 
+To work out our strategy, let's consider what we already know how to do. We already know how to take a key-value pair as a string, and add it to our object.
+
 > 💡 Key insight: If we can do it for **one pair**, we can try doing it for a **list of pairs** too.
 
+So we're missing a step - breaking up the string of lot sof key-value pairs into an array where each element is a single key-value pair. If we do this, then we can iterate over the array, and do what we already know how to do on each key-value pair.
```
Dedekind561, line 35:
```suggestion
So we're missing a step - breaking up the string of lots of key-value pairs into an array where each element is a single key-value pair. If we do this, then we can iterate over the array, and do what we already know how to do on each key-value pair.
```

content/js2/blocks/multiple-params/index.md
```diff
@@ -74,3 +78,28 @@ function parseQueryString(queryString) {
 {{</tab>}}
 
 {{</tabs>}}
+
+Now that we've worked out how to solve this problem in the case of multiple query parameters, let's integrate that solution into our previous implementation, to make sure it works for all cases.
+
+We can keep our `if (queryString.length === 0) {` check from before. We don't need to do anything special for the one-value case, as an array containing one element gets iterated the same as an array multiple elements:
```
Dedekind561, line 84:
```suggestion
We can keep our `if (queryString.length === 0) {` check from before. We don't need to do anything special for the one-value case, as an array containing one element gets iterated the same as an array of multiple elements:
```

content/js2/blocks/no-params/index.md
```diff
@@ -37,6 +43,8 @@ However, after re-running the tests, we get the following feedback:
 
 ### Checking objects
 
+We saw earlier that objects, like arrays, are reference types. That means that comparing two objects with `===` will only evalute to `true` if they are references to the same objects in memory. Two objects which happen to have the same properties, but are not in the same memory, will compare `false` using `===`.
```
Dedekind561, line 46:
```suggestion
We saw earlier that objects, like arrays, are reference types. That means that comparing two objects with `===` will only evalute to `true` if they are references to the same objects in memory. Two objects which happen to have the same properties, but are not in the same location in memory, will compare `false` using `===`.
```

content/js2/blocks/query-string/index.md
```diff
@@ -7,18 +7,18 @@ emoji= '🧩'
 [objectives]
 1='Identify the query string section of a url'
 2='Identify query parameters within a query string'
-3='Explain why an object is a more useful structure for storing query parameters'
+3='Explain why an object is a more useful structure for storing query parameters than'
```
Dedekind561, line 10:
```suggestion
3='Explain why an object is a more useful structure for storing query parameters than a string' 
```

content/js2/blocks/multiple-params/index.md
```diff
@@ -42,19 +46,19 @@ Query strings with multiple key-value pairs use `&` as a separator e.g. `sort=lo
...
   const queryParams = {};
-  const keyValuePairs = queryString.split("&"); // keyValuePairs will point to [ "sort=yellow", "colour=yellow"]
+  const keyValuePairs = queryString.split("&"); // keyValuePairs will point to ["sort=yellow", "colour=yellow"]
 }
 ```
 
 #### 🎯 Sub-goal 2: add each key-value pair in the array to the query params object
 
-Once we've got an array we can iterate through the key-value pairs and update the `queryParams` object each time.
+Once we've got an array we can iterate through the key-value pairs and update the `queryParams` object each time (just like we did when we just had one key-value pair).
```
Dedekind561, line 55:
```suggestion
Once we've got an array we can iterate through the key-value pairs and update the `queryParams` object each time (like we did when we just had one key-value pair).
```

content/js2/blocks/query-string/index.md
```diff
@@ -7,18 +7,18 @@ emoji= '🧩'
...
+Websites have addresses called urls like this: "https://example.com/widgets". Urls often have {{<tooltip title="query strings">}}Query strings go at the end of a url and are used to specify more information about the content you get back from a request to a server{{</tooltip>}}too. Here is an example of a url with a query string on the end:
 
 https://example.com/widgets?colour=blue&sort=newest
 
 For the url above, the **query string** is `"colour=blue&sort=newest"`. Query strings consist of **query parameters**, separated by an ampersand character `&`. `colour=blue` is a query parameter: we say that `colour` is the key and `blue` is the value.
 
-urls must always be strings. However, a string isn't a useful data type for accessing query parameters. Given a key, accessing the value from a query string is not straightforward However, objects are ideal for looking up values with keys.
+urls must always be strings. However, a string isn't a useful data type for accessing query parameters. Given a key like `colour`, accessing the value from a query string stored as a string is not straightforward. However, objects are ideal for looking up values with keys.
```
Dedekind561, line 21: Accessing the value from a query string stored as a string doesn't scan well - can we leave it as query string?

content/js2/blocks/variable-keys/index.md
```diff
@@ -39,6 +39,8 @@ function parseQueryString(queryString) {
 }
 ```
 
+We can't use dot syntax if we don't know when we're writing the code what the name of the key is going to be. Square bracket notation is more powerful than dot notation, because it lets us use variables as keys.
```
Dedekind561, line 42:
```suggestion
We can't use dot syntax if we don't know what the name of the key is going to be. Square bracket notation is more powerful than dot notation, because it lets us use variables as keys.
```

content/js2/blocks/multiple-params/index.md
Dedekind561: I'm afraid of there being too much content in the prep. Can we move all the edge case material into a separate exercise in the coursework? Or even do a workshop to explore edge cases and how to handle them?

content/js2/blocks/query-string/index.md
illicitonion: Works for me :)

content/js2/blocks/multiple-params/index.md
illicitonion: I'm happy to skip the exercise, but I'd like to keep the prompt to think about them... What do you think?

Dedekind561: Yes absolutely - I like the prompt

PR #274, opened 2023-09-10
Underline tooltip text

illicitonion:
It wasn't super obvious to me at a glance that these were hoverable - this maybe makes that more obvious.

### DIFF
--- a/assets/styles/04-components/tooltip.scss
+++ b/assets/styles/04-components/tooltip.scss
@@ -2,6 +2,7 @@ tool-tip-container {
   position: relative;
   display: inline-block;
   cursor: help;
+  text-decoration: dashed underline;
 }
 
 .tooltip {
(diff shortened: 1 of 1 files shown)


assets/styles/04-components/tooltip.scss
```diff
@@ -2,6 +2,7 @@ tool-tip-container {
   position: relative;
   display: inline-block;
   cursor: help;
+  text-decoration: dashed underline;
```
SallyMcGrath, line 5:
```suggestion
  text-decoration: 1px var(--theme-color--brand) wavy underline;
```

A bit tight and heavy as it was. How about this?

SallyMcGrath (changes requested):
Don't object to the idea - how about this to connect it to the ball of yarn a little better?

assets/styles/04-components/tooltip.scss
illicitonion: Love it! Thanks!

SallyMcGrath (approved):
LGTM

PR #291, opened 2023-09-16
Handle empty PR response by hiding component

illicitonion:
Also, fix the error message


40thieves (approved):
Good detective work! 🕵️‍♂️

PR #297, opened 2023-09-20
Make example code valid

illicitonion:
## What does this change?

Module: JS1
Week(s): 2

## Description

I don't think it's super clear that trainees are expected to put in placeholder values, and the SyntaxError they'll get if they copy + paste the code as-is doesn't seem to have great learning value.


Dedekind561 (approved):
LGTM

PR #309, opened 2023-09-30
added some extra instructions for connecting to render postgres from render

jxz12:
since many trainees struggled with it (https://codeyourfuture.slack.com/archives/C0452U6V0UX/p1694639075135679)

## What does this change?

Module: SQL
Week(s): 4(?)

## Checklist

## Description

Added an extra section to the `Connecting a client to a database on Render` guide since many trainees spent a lot of time trying to get it working

@kfklein15 @illicitonion


jxz12:
One thing I'm not sure of is linking to another markdown file via `[integration with node](../../../databases/blocks/integration-with-node/index.md)`

If someone else has `hugo` installed could they check if it works please? Otherwise I will try to install it locally :)

kfklein15 (approved):
Clear instructions.

jxz12:
Thank you for the review @kfklein15!

Actually, after seeing the cool netlify deployment I don't think the markdown link works (`[integration with node](../../../databases/blocks/integration-with-node/index.md)`), so I will fix/remove it before merging!

jxz12:
I couldn't work out how to reference another block so just removed the link instead, so I think this PR is ready!

Another trainee also mentioned that they struggled with environment variables (which can be used to not hard code passwords in git) so I will also make a PR to add instructions for that too.

kfklein15 (approved):
Looks good to me. Hopefully it will make things clearer for everyone.

illicitonion (approved):
LGTM, thanks!

I'll merge this because it's definitely better to have than not, but also ask two follow-ups questions:
1. Do we want to just always suggest using a `connectionString` rather than the individual fields? I believe the external database URL should _always_ work, and it's less hassle for folks to have to `.env`-ify one variable rather than 4
2. Do we want to mention `.env` files here at all? In the full-stack assessment, 10-20% of our trainees checked in their raw DB credentials in their JS (and another ~10% checked in their `.env` file, but that's another story :D). One of the things I saw most of my project pathway folks struggle with was they didn't realise on Render there's an "Environment" tab for a backend where they can set environment variables that will be used equivalent to using a `.env` file... (I can buy the argument this is a separate topic, though)

PR #330, opened 2023-10-22
Edit feedback on JS2W3 rewrite

illicitonion:


content/js2/blocks/events/index.md
```diff
@@ -22,11 +22,17 @@ However, we're missing a step in our plan. We need to find a way of running some
...
 {{</note>}}
 
-Events are things that happen in the browser, which the browser tells you about so your code can react to them. In a browser context, events could be a user clicking on a button, a user typing something into a textarea box, submitting a form etc. You can find a [complete reference all the different event types](https://developer.mozilla.org/en-US/docs/Web/Events) on MDN.
+Events are things that happen in the browser, which your code can ask to be told about, so that your code can react to them. In a browser context, an event could be a user clicking on a button, a user typing something into a textarea box, a user submitting a form etc. Not all events are in response to user actions, for instance there is an event for the browser finished initially rendering the page. You can find a [complete reference all the different event types](https://developer.mozilla.org/en-US/docs/Web/Events) on MDN.
 
-In the case of the `textarea` element, there are lots of events occurring. The `textarea` value is changing, a keyboard button is released every time the user types and so on. We can update our plan as follows:
```
illicitonion, line 27: I don't think it was salient here to describe that a key-press triggers multiple events, it felt like a distraction. It's definitely interesting and useful, but probably not required here?

content/js2/blocks/response/index.md
```diff
@@ -31,26 +31,26 @@ textarea.addEventListener("keyup", handleKeyup);
...
+- [`"keyup"`](https://developer.mozilla.org/en-US/docs/Web/API/Element/keyup_event) - this is the type of event we want to be notified about
+- `updateCharacterLimit` - the second argument is a function. It is a function that is called when an event occurs.
 
-In JavaScript, we can pass functions as arguments to other functions. In this case, we're passing a function `handleKeyup` to `addEventListener` as an input. We can think of this as saying: whenever a key is released on the `textarea` element, then the function `handleKeyup` will get called. Any code we want to run in response to the `keyup` event will need to be executed inside `handleKeyup`.
+In JavaScript, we can pass functions as arguments to other functions. In this case, we're passing a function `updateCharacterLimit` to `addEventListener` as an input. We can think of this as saying: whenever a key is released on the `textarea` element, then the browser will call the function `updateCharacterLimit`. Any code we want to run in response to the `keyup` event will need to be inside the `updateCharacterLimit` function.
```
illicitonion, line 37: When were you planning on introducing `Array.map` and `Array.filter`? I was going to make a comparison here, but realised we haven't actually covered them yet. I don't think that's a problem, but it triggered the question for me :)

illicitonion, line 37: I think being clear with the subject in active language here is useful, rather than vaguely/passively saying "will be called".

content/js2/blocks/response/index.md
```diff
@@ -11,46 +11,46 @@ emoji= '🧩'
 
 As a user, we interact with the elements on a web page. We _click_ on buttons, _input_ text, _submit_ forms etc.
 
-To react to an event, we can declare a function that we want to run whenever a certain event occurs. We call this function an **event handler**. In the example below, we _name_ this function `handleKeyup`:
```
illicitonion, line 14: I want to specifically _not_ use `handleKeyup` as a name here, because people start thinking they're magical names (either that the name must be exactly that, or that they shouldn't think of better names because there's an "obvious" name).

Let's establish this practice early :)

content/js2/blocks/response/index.md
illicitonion, line 1: I can't comment on the line, but:

Is this the first time we've introduced the "Console" view of the web inspector?

I suspect it is, because the trainees haven't run JS in the browser before.

If so, we probably need some screenshots and explanation of this surface?

content/js2/blocks/response/index.md
Dedekind561: I'm thinking that week 1 of JS3 will be the ideal place to introduce `map` as we're going to talk about mapping a list of data to a list of UI elements.

content/js2/blocks/response/index.md
Dedekind561: Absolutely!

content/js2/blocks/response/index.md
illicitonion: Sounds good!

content/js2/blocks/response/index.md
illicitonion: Agree it doesn't block the PR - I think ideally we probably need to put together both:
1. A guide on https://curriculum.codeyourfuture.io/guides/
2. A workshop we can do in-class

Which should cover both the console, and also the fact that the element tree reflects the underlying DOM rather than original HTML (and so we can see changes from our JS in it)

illicitonion: Filed https://github.com/CodeYourFuture/curriculum/issues/331 to track this

Issue #333, opened 2023-10-25
Issues in module repository backlog do not appear on the curriculum site's backlog


JayMayer:
**Describe the bug**

GitHub issues from module repos are not appearing on the curriculum site's backlog.

**What is wrong with this content?**

Trainees and volunteers are unable to get a complete picture of the backlog from the curriculum site.

**How could this be fixed?**

Ensure that the backlog displayed on the curriculum site is the same as the backlog in the module repository.

**Link to content**

Example using HTML-CSS Sprint 1 (note the content is different)
- [Curriculum Site Backlog](https://curriculum.codeyourfuture.io/html-css/sprints/1/backlog/)
- [Module Repo Backlog](https://github.com/orgs/CodeYourFuture/projects/87/views/4?filterQuery=-status%3A%22%E2%9C%85+Done%22+week%3A%22Week+1%22)

**Additional context**

Not all issues are affected, some are displayed.

The bug has only been observed recently, in the past all issues have been displayed.

**Who else might need to know about this?**

@AltomHussain @SallyMcGrath

SallyMcGrath:
OK walking through this. I've spun up the repo and running it on local. 

On 

http://localhost:1313/html-css/sprints/1/backlog/ = 2 issues showing of 
http://localhost:1313/html-css/sprints/3/backlog/ = 3
http://localhost:1313/html-css/sprints/5/backlog/ = 5

the issues api being passed to the issues partial on all these views is https://api.github.com/repos/CodeYourFuture/Module-HTML-CSS/issues?per_page=100 and all the issues show on this view

I move into the partial. Inside the issues range, the first piece of logic is filtering the issues to show if a label matches the parameter passed in. I comment this out, to see if it shows all the issues. I get a giant error

[...]

PR #335, opened 2023-10-26
Bugfix warn on empty markdown

SallyMcGrath:
Addresses #333

I haven't put it as an error (which will fail the build) because the nature of this platform is that we will be pulling in UGC from many places and we won't have control of all the markdown. I don't want to throw errors a contributor cannot fix.

You can't exit a render hook with a warning, so I've placed a check at the start for empty .Destination and printed a warning to the console.

## test

Ive created a broken image in an issue here
https://github.com/CodeYourFuture/Module-Template/issues/6

You can pull it in as a block to see the page build with no image (not even broken in this case, which we can argue about if you like...)

```
[[blocks]]
name="MRE broken image"
src="https://github.com/CodeYourFuture/Module-Template/issues/6"
```

### DIFF
--- a/layouts/_default/_markup/render-image.html
+++ b/layouts/_default/_markup/render-image.html
@@ -1,81 +1,85 @@
-{{ $src := .Destination }}
-{{ $alt := .PlainText | safeHTML }}
-{{ $caption := "" }}
-{{ with .Title }}
-  {{ $caption = . | safeHTML }}
-{{ end }}
-
-{{/* Check if the source is an external link */}}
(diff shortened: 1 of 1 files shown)


SallyMcGrath:
Ugh, I don't know why it's reformatted the whole block -- I do have linting in this repo!

```
{{ if eq .Destination "" }}
  {{ warnf "Image source is empty: %s" .PlainText }}
```
  
  was my change

Dedekind561 (approved):
LGTM, I get the correct warning when firing up the development environment. Added that block to the one the day plans to double-check it in the UI.

PR #336, opened 2023-10-26
scope css to wordlimit

Dedekind561:
## What does this change?

- Scope CSS to word limit component only


layouts/shortcodes/wordlimit.html
```diff
@@ -1,10 +1,10 @@
-<section>
+<section id="wordlimit">
```
SallyMcGrath, line 1: If this is a shortcode, presumably it can be called on a page mulitple times. In which case, please uniquify the id.

Secondly, please add a BEM class and put your styles on the class not the id. TY!

assets/styles/04-components/wordlimit.scss
```diff
@@ -1,8 +1,8 @@
-section {
+#wordlimit {
   display: grid;
   width: 50%;
 }
 
-label {
+#wordlimit > label {
   margin-bottom: 1em;
```
SallyMcGrath, line 7:
```suggestion
  margin-bottom: var(--theme-spacing--gutter);
```

Please don't use literals - use the existing design tokens.

layouts/shortcodes/wordlimit.html
SallyMcGrath, line 1: If this is a shortcode, presumably it can be called on a page mulitple times. In which case, please uniquify the id.

Secondly, please add a BEM class and put your styles on the class not the id. TY!

SallyMcGrath (changes requested):
A few requests ty!

Dedekind561:
Hi there, 
- [x] I've refactored this into web component after looking at some of the other web components that make use of event listeners so it seemed like putting the word count logic into a web component makes more sense. Ive then called the web component from the short code.
- [x] I've applied the CSS to the shadow element and it's not leaking out into the main page like before.
- [x] I've used bem class names for the elements

Dedekind561:
Brief update:
I updated the branch this morning as the shortcode was swallowing up some of the shadow content from other shortcodes being used in the `character-limit` block in JS2. There's just one thing I'm struggling with: how to _uniquify_ id attributes for HTML elements created in a web component. I had a quick search online but couldn't see anything particularly useful.

SallyMcGrath:
> Brief update: I updated the branch this morning as the shortcode ( `wordlimit` ) was swallowing up some of the shadow DOM content from other shortcodes used in the `character-limit` block in JS2 week 3 prep view. There's just one thing I'm struggling with: how to _uniquify_ id attributes for HTML elements created in a web component. I had a quick search online but couldn't see anything particularly useful.

If it's in a web component it's in the shadow so you don't need to worry about it. The shortcode would be compiled into the main document. If it's all in the same document, the simplest way to uniquify the id is to append a random number or EPOCH number to the name of the id, like `{{now.UnixNano}}`

SallyMcGrath:
Screenshot 2023-10-28 at 19 29 47


I am really bothered by the tiny boxes and I'm going to do a ton of content population so I'm going to merge this and we'll look again in the week.

SallyMcGrath (approved):

PR #366, opened 2023-11-04
JS1 Week 2 tweaks/improvements

40thieves:
## What does this change?

Module: JS1
Week(s): 2

## Checklist

## Description

While reading through the new content (I'm trying to catch up again), I spotted some issues:

- In the Percentages section, there's a very abrupt change from explanatory content to (what I assume is) practice "exercises". I've added a bit more context and reworded to be a bit less maths-y
- In the Returning from a function section, I thought it was a bit odd to show the code snippet directly returning the expression first, instead of returning the variable that trainees would have already seen. So I switched those around
- At the end of the Returning from a function section, I spotted a bug in the code snippets that show logging of the output (return statements missing). I also added an example of what the output should be.

I'm happy to discuss any changes to this/revert anything.

### DIFF
--- a/content/js1/blocks/percentages/index.md
+++ b/content/js1/blocks/percentages/index.md
@@ -8,7 +8,7 @@ emoji= '🗄️'
     2='Construct simple expressions to represent a calculation'
 +++
 
-Let's begin with this problem.
+Let's begin with this problem:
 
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
LGTM, thanks!

PR #392, opened 2023-11-24
List issues via a pagination-expanding proxy

illicitonion:
Previously we were just missing any issues that didn't happen to
paginate on the first page.

For JS3, this was quite a few issues.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -27,6 +27,9 @@ tree = "https://github.com/CodeYourFuture/curriculum/tree/main/"
 edit = "https://github.com/CodeYourFuture/curriculum/edit/main/"
 root = "curriculum"
 orgapi = "https://api.github.com/repos/CodeYourFuture/"
+# We use a proxy which concatenates paginated responses, otherwise we miss results.
+# Daniel is currently hosting this code https://github.com/illicitonion/github-issue-proxy but we should work out a long-term maintainable solution to this problem.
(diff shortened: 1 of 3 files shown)


SallyMcGrath:
Oh yes I hit this before too. I guess when I hit it I was thinking do we want that many issues anyway? 🤔 

You have access to our Netlify if you want to put a serverless function on there btw. As you have presumably already read, GitHub [expects us to use Octokit.js for this ](https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28). I'm happy to see Rust though! What do you want to do long term?

Thanks also for tidying up my variable names which were obviously copypastad from some other use case. ❤️

SallyMcGrath (approved):

illicitonion:
> I guess when I hit it I was thinking do we want that many issues anyway?

I think the problem is that we're filtering on labels _after_ the fetch rather than in the query, which means we don't have as much control over how many issues this means we have.

> You have access to our Netlify if you want to put a serverless function on there btw. As you have presumably already read, GitHub [expects us to use Octokit.js for this ](https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28). I'm happy to see Rust though! What do you want to do long term?

Yeah, a JS/TS serverless function on Netlify probably makes sense... I filed https://github.com/CodeYourFuture/curriculum/issues/393 for us to get back to this at some point, but I'm not going to pick it up right now.

PR #444, opened 2023-12-13
Implement dev mode for faster iteration

### LINKED ISSUE #431: Set up a dev mode and put caching on

Stuff like getting paginated issues slows us down and is rarely helpful in dev.

I'd like a speedier devmode which doesn't constantly pull from the apis. Ideally one I can run offline on planes!

`hugo.IsProduction` could help maybe?

**Additional context**

illicitonion:
Fixes #431


SallyMcGrath (approved):
YES 1715ms

PR #458, opened 2023-12-15
Feature: Runkit -- I want executable runkits for Node

SallyMcGrath:
## What does this change?

Module: Node + all
Week(s): all

## Checklist

## Description

The Express docs use this and it's pretty easy to create notebooks directly from a render hook.

https://runkit.com/home

It's not themeable, really, so far as I can see. Sadface. But you can do this:

- all documents on RunKit are public, you can make them searchable by publishing.
- all of npm's 1,000,000+ packages are already pre-installed, just require() them
- use arrow functions, classes, template strings, and most of ES6. See everything we support here.
- await any promise instead of using callbacks (example)
- store secrets in environment variables
- export an endpoint function to turn any notebook into an API (endpoint docs)

# Usage - creates the runkit directly from the code block

```runkit
//Load the 'express' library to easily handle HTTP conversations:
const express = require("express");

const app = express();

//Register some handlers for different routes.
app.get("/", function (request, response) {
    response.send("Hello CYF");
});
app.get("/two", function (request, response) {
    response.send("Another route");
});
app.get("/numbers", function(request, response) {
  const someData = [1, 2, 3];
  response.json(someData);
});
//Tell the server to start listening for requests.
app.listen(3000);
```

### DIFF
--- a/assets/styles/03-elements/table.scss
+++ b/assets/styles/03-elements/table.scss
@@ -8,6 +8,10 @@ table {
   border-image: var(--theme-border-image);
 }
 
+.highlight table {
+  width: 100%;
+}
+
 thead th {
   font: var(--theme-font--system);
   background-color: var(--theme-color--block);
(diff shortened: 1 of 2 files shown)


40thieves (approved):
I wasn't able to test this locally (mostly out of curiosity about how it looked) - but it seems there's an issue with that at the moment.

I don't see how this PR would cause issues though, so happy to ship it.

PR #462, opened 2023-12-18
Copy edits to JS3W3

illicitonion:


content/js3/blocks/using-fetch/index.md
```diff
@@ -23,11 +23,13 @@ But we still don't know how to use `fetch` to get data from a server side API. L
 ```js
 const endpoint = "//curriculum.codeyourfuture.io/dummy-apis/films.json";
 const fetchPromise = fetch(endpoint);
-fetchPromise.then((response) => {
+const _ = fetchPromise.then((response) => {
```
illicitonion, line 26: If you don't assign this to something, you end up with _two_ promises `console.log`'d, which I think is confusing rather than clear at this point...

PR #473, opened 2023-12-20
do not overwrite baseURL

SallyMcGrath:
we don't need to redirect inside the language unless we are producing sites
so I've removed the config of baseURL inside the /en config 

so assets will work again on your local


40thieves (approved):
It works 🎉 Thanks!

PR #474, opened 2023-12-21
Revise JS3 1 prep - template tags

SallyMcGrath:
## What does this change?

Module: JS3
Week(s): 1

## Checklist

- [ ] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend

## Description

Reading JS3 1 and the new revisions have some errors
I have fixed them up I think and also cut a bit more to simplify

### DIFF
--- a/content/en/js3/blocks/components/index.md
+++ b/content/en/js3/blocks/components/index.md
@@ -14,7 +14,6 @@ Recall our sub-goal:
 
 Now that we have made a card work for one particular film, we can re-use that code to render _any_ film object in the user interface with a general component. To do this, we wrap up our code inside a JavaScript function. JavaScript functions **reuse code**: so we can implement **reusable UI components** using functions.
 
-
 ```js
 const film = {
   title: "Killing of Flower Moon",
@@ -27,22 +26,11 @@ const film = {
 const createFilmCard = (film) => {
   const card = document.getElementById("film-card").content.cloneNode(true);
   // Now we are querying our cloned fragment, not the entire page.
-  card.querySelector('h3[name="title"]').textContent = film.title;
-
-  card.querySelector('p[name="director"]').textContent = `Director: ${film.director}`;
-
-  const times = document.querySelector('ul[name="times"]');
-  for (const time of film.times) {
-    // We could use either createChildElement, or a different <template> tag here if we wanted.
-    const timeElement = document.createElement("li");
-    timeElement.textContent = time;
-    times.append(timeElement);
-  }
-
-  card.querySelector('p[name="duration"]').textContent = `${film.duration} minutes`;
-
-  card.querySelector('p[name="certificate"]').textContent = `Certificate: ${film.certificate}`;
-  // Return the card, rather than directly appending it to the page, because it means the caller could do something else with it if they wanted. e.g. if we wanted to show cards in different parent elements, we could still re-use this function.
+  card.querySelector("h3").textContent = film.title;
+  card.querySelector("p").textContent = `Director: ${film.director}`;
+  card.querySelector("time").textContent = `${film.duration} minutes`;
+  card.querySelector("data").textContent = `Certificate: ${film.certificate}`;
+  // Return the card, rather than directly appending it to the page
   return card;
 };
 const template = document.getElementById("filmCard");
--- a/content/en/js3/blocks/simplifying-element-creation/index.md
+++ b/content/en/js3/blocks/simplifying-element-creation/index.md
@@ -6,7 +6,6 @@ facilitation = false
 emoji= '🧩'
 [objectives]
     1='Extract functions for common tasks'
-    2='Use template tags to simplify element initialisation'
 +++
 
 We now have a card showing all of the information for one film.
@@ -26,6 +25,7 @@ To refactor means to update our code _quality_ without changing the _implementat
 We can identify things we're doing several times, and extract a function to do that thing for us.
 
 In this example, we keep doing these three things:
+
 1. Create a new element (sometimes with a different tag name).
 2. Set that element's text content (always to different values).
 3. Appending that element to some parent element (sometimes a different parent).
@@ -36,10 +36,10 @@ We could write a function like this:
 
 ```js
 function createChildElement(parentElement, tagName, textContent) {
-    const element = document.createElement(tagName);
-    element.textContent = textContent;
-    parentElement.append(element);
-    return element;
+  const element = document.createElement(tagName);
+  element.textContent = textContent;
+  parentElement.append(element);
+  return element;
 }
 ```
 
@@ -55,10 +55,10 @@ const film = {
 };
 
 function createChildElement(parentElement, tagName, textContent) {
-    const element = document.createElement(tagName);
-    element.textContent = textContent;
-    parentElement.append(element);
-    return element;
+  const element = document.createElement(tagName);
+  element.textContent = textContent;
+  parentElement.append(element);
+  return element;
 }
 
 const card = document.createElement("section");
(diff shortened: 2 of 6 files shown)


content/en/js3/blocks/components/index.md
```diff
@@ -27,22 +26,11 @@ const film = {
 const createFilmCard = (film) => {
   const card = document.getElementById("film-card").content.cloneNode(true);
   // Now we are querying our cloned fragment, not the entire page.
-  card.querySelector('h3[name="title"]').textContent = film.title;
-
-  card.querySelector('p[name="director"]').textContent = `Director: ${film.director}`;
-
-  const times = document.querySelector('ul[name="times"]');
-  for (const time of film.times) {
```
illicitonion, line 35: I think the "times" loop here is actually the most interesting thing in this example, mostly because it shows that even when using `template` sometimes it makes sense to mix in raw DOM APIs - what's your motivation in removing it?

content/en/js3/blocks/simplifying-element-creation/index.md
```diff
@@ -67,100 +65,30 @@ createChildElement(card, "h3", film.title);
...
-The first two approaches (the verbose version, and the `createChildElement` version) did so by calling the same DOM functions as each other.
+1. If we want to change how we create some, but not all, elements, we may have made it _harder_ to make these changes. When we want to include an image of the director, or replace the certificate text with a symbol, we will have to introduce branching logic.
+2. To follow how something is rendered, we need to look in a few places. This is something you will need to get used to, so it's good to start practising now.
 
-This approach uses _different_ DOM functions. But it has the same effect.
+{{<note type="exercise" title="Deep dive: datetime">}}
 
-{{<note type="exercise" title="Exercise: Consider the trade-offs">}}
-We've now seen two different ways of simplifying our function: extracting a function, or using a template tag.
+### Stretch goal: Add the datetime
```
illicitonion, line 88: I like this, it's a nice little exercise!

content/en/js3/blocks/template-html/index.md
```diff
@@ -0,0 +1,69 @@
...
+
+We could simplify this code with a different technique for creating elements.
+
+Until now, we have only seen one way to create elements: `document.createElement`. The DOM has another way of creating elements - we can _copy_ existing elements and then change them.
+
+HTML has a useful tag designed to help make this easy, [the `<template>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template). When you add a `<template>` element to a page, it doesn't get displayed when the page loads. It is an inert fragment of future HTML.
+
+We can copy any DOM node, not just `<template>` tags. For this problem, we will use a `<template>` tag because it is designed for this purpose.
+
+When we copy an element, its children get copied. This means we can write our template card as HTML:
```
illicitonion, line 21: One thing I realise we skip over is that `<template>`, unlike other elements, strips way its `<template>` wrapper when we clone it (which is one of its super powers!) - we should maybe call this out explicitly as a difference...

content/en/js3/blocks/template-html/index.md
```diff
@@ -0,0 +1,69 @@
...
+  <section>
+    <h3>Film title</h3>
+    <p>Director</p>
+    <time>Duration</time>
+    <data>Certificate</data>
+  </section>
+</template>
+```
+
+This is our template card. Place it in the body of your html. It doesn't show up! [Template HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) is like a wireframe; it's just a _plan_. We can use this template to create a card for any film object. We will clone (copy) this template and populate it with data.
```
illicitonion, line 34: This is a good call-out :)

content/en/js3/blocks/template-html/index.md
```diff
@@ -0,0 +1,69 @@
...
+  director: "Martin Scorsese",
+  times: ["15:35"],
+  certificate: "15",
+  duration: 112,
+};
+
+const card = document.getElementById("film-card").content.cloneNode(true);
+// Now we are querying our cloned fragment, not the entire page.
+card.querySelector("h3").textContent = film.title;
+card.querySelector("p").textContent = `Director: ${film.director}`;
```
illicitonion, line 48: I don't love the assumption that there is exactly one `p` tag and it's for the director - this feels like one where I would want a `name` or `class` attribute included in the query selector, WDYT?

content/en/js3/blocks/template-html/index.md
```diff
@@ -0,0 +1,69 @@
...
+card.querySelector("time").textContent = `${film.duration} minutes`;
+card.querySelector("data").textContent = film.certificate;
```
illicitonion, line 50: Should a `data` tag not also have a `value` attribute? What's the advantage of using a `<data>` tag over a `<p>` tag here?

illicitonion (commented):
Thanks for the clean-up, this looks good!

A few questions in comments :)

content/en/js3/blocks/components/index.md
SallyMcGrath: Oh one conversation in three places!!

For here, I paste for anyone interested

So, usually you do end up with a separate time creating function when building these cards. (I worked for 3 years building cinema and theatre listings.)  This is because handling time and making them accessible and internationalisable usually includes a bunch of other stuff as well. That's why I removed the list creation. It isn't how you would handle that issue. 

I could have converted it, like this, by putting a time template node in and then cloning and appending that
```
const timeTemplateNode = card.querySelector("time");
        times.map((time) => {
          const timeClone = timeTemplateNode.cloneNode();
          timeClone.textContent = time;
          timeTemplateNode.parentNode.appendChild(timeClone);
        });
        timeTemplateNode.remove();

```

or whatever. But that's not what you would really do with times, so I felt like it was too artificial and just bulking out the code samples  -- as yours was too and the bulk I felt was confounding and not clarifying.


Re the name attribute I removed: the attribute you would usually use to identify is data-
so data-director, data-times, data-certificate

is that part useful to wedge back in? Showing how to handle the same elements with different identifiers?

content/en/js3/blocks/template-html/index.md
SallyMcGrath: Re the name attribute I removed: the attribute you would usually use to identify is data-
so data-director, data-times, data-certificate. Name isn't a valid attribute on p. 

is that part useful to wedge back in? Showing how to handle the same elements with different identifiers? In almost all cases, the selector should be either the html element or a data-attribute. Is it too much HTML lore though?

content/en/js3/blocks/template-html/index.md
SallyMcGrath: Yeah it does. We often used data tags for certificates. Data is just like the time tag -- it gives you a human readable version and a structured data value, which is nice for thinking about different consumers of the same data.  I am torn here -- I just think the code samples are getting too long.

content/en/js3/blocks/components/index.md
illicitonion: Sounds good - thanks!

content/en/js3/blocks/template-html/index.md
illicitonion: Yeah, I think it'd be good to include a more semantic selector in there - I don't think we need to really talk about it (though I guess we can somewhere - maybe way back in HTML/CSS), but I think it'd be good to improve that example slightly :)

SallyMcGrath: https://github.com/CodeYourFuture/curriculum/pull/474/commits/ee6e8d0ffd610d433e91f8a2ae5beb32f85a4891

illicitonion (approved):

PR #484, opened 2023-12-30
JS3 Sprint 4 prep: Promises, then, async/await, try/catch, getFilms

SallyMcGrath:
## What does this change?

Module: JS3
Week(s): 4

## Checklist

## Description

Completing the concept map from JS3-3 with the final parts on Promises, then, async/await, and try/catch. I brought filterFilms back from week 2 and updated it with a fetch as the last part.

Please do checkout this PR and edit directly as well as offering line edits. If we can simplify or reduce, let's do this!

I will go away and fix the horrible Mermaid in solo view bug after React is ported. For now just reload the page (sorry!!!)

### DIFF
--- a/content/en/js3/blocks/async-await/index.md
+++ b/content/en/js3/blocks/async-await/index.md
@@ -1,11 +1,33 @@
 +++
-title = 'block3'
+title = 'Async/Await'
 headless = true
 time = 30
 facilitation = false
 emoji= '🧩'
 [objectives]
-    1='Use the Teach Tech Together guide to construct your objectives'
-    2='Limit the objectives to 3-5 items'
-    3='Write objectives you can measure'
+    1='Define syntactic sugar'
+    2='Write a function using the async keyword'
 +++
+
+Async/await is {{<tooltip title="syntactic sugar">}}A simpler, or "sweeter" way to write the same thing. The code works the same under the hood, but it's easier to read. {{</tooltip>}} for Promises. We group them together: async/await, because we {{<tooltip title="use them together. ">}}We can only use `await` inside an `async` function or at the [top level](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) of a module.{{</tooltip>}}
+
+We use the `async` [keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords) to define a function that returns a Promise.
+
+```js
+const getProfile = async (url) => {
+  // the async keyword tells us this function handles a time problem
+};
+```
+
+We use the [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) operator to wait for a Promise to resolve. This allows us to write code that looks like it's happening in time order, but doesn't block our main thread.
+
+```js
+const getProfile = async (url) => {
+  const response = await fetch(url);
+  return response.json();
+};
+```
+
+When we use `await`, we are saying, "Wait for this Promise to resolve before moving on to the next line of code." But if the Promise _doesn't_ resolve, the next line of code will never run and an [error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) will be thrown. But how can we find out what the error is?
+
+We need to `catch` the error.
--- a/content/en/js3/blocks/async-await/index.md
+++ b/content/en/js3/blocks/async-await/index.md
@@ -20,6 +20,20 @@ Async/await is {{<tooltip title="syntactic sugar">}}A simpler, or "sweeter" way 
 
 We use the [`async`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) [keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords) to define a function that returns a Promise. An async function always returns a Promise.
 
+We can see this with a simple function which doesn't need to await anything:
+
+```js
+const getProfile = async (url) => url;
+
+console.log(getProfile("hello")); // Logs a Promise.
+
+getProfile("hello").then((value) => console.log(value)); // Logs a value
+```
+
+Even though the function above doesn't _have_ a time problem, the fact that we define the function as an `async` function means it returns a `Promise`.
+
+But let's do something more interesting - let's actually solve a time problem.
+
 ```js
 const getProfile = async (url) => {
   // the async keyword tells us this function handles a time problem
@@ -35,7 +49,7 @@ const getProfile = async (url) => {
 };
 ```
 
-Go ahead and call this in your Node REPL in your terminal: `getProfile("https://api.github.com/users/SallyMcGrath")`. It works the same as before.
+Go ahead and call this in your Node REPL in your terminal: `getProfile("https://api.github.com/users/SallyMcGrath").then(console.log)`. It works the same as before.
 
 ### 🫠 Handling errors
 
--- a/content/en/js3/blocks/async-await/index.md
+++ b/content/en/js3/blocks/async-await/index.md
@@ -1,7 +1,7 @@
 +++
-title = 'Async/Await'
+title = '🍬 async/await'
 headless = true
-time = 30
+time = 20
 facilitation = false
 emoji= '🧩'
 [objectives]
@@ -9,9 +9,16 @@ emoji= '🧩'
     2='Write a function using the async keyword'
 +++
 
+```mermaid
+graph LR
+    Promise{{🤝 Promises}} --> |syntax| async{{🏃‍♂️ async}}
+    async --> |syntax| await{{📭 await}}
+    await --> |resolves to| Response{{📤 Response}}
+```
+
 Async/await is {{<tooltip title="syntactic sugar">}}A simpler, or "sweeter" way to write the same thing. The code works the same under the hood, but it's easier to read. {{</tooltip>}} for Promises. We group them together: async/await, because we {{<tooltip title="use them together. ">}}We can only use `await` inside an `async` function or at the [top level](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) of a module.{{</tooltip>}}
 
-We use the `async` [keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords) to define a function that returns a Promise.
+We use the [`async`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) [keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords) to define a function that returns a Promise. An async function always returns a Promise.
 
 ```js
 const getProfile = async (url) => {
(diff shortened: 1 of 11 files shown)


content/en/js3/blocks/promises/index.md
```diff
@@ -0,0 +1,67 @@
...
+```mermaid
+graph LR
+    Asynchrony --> |delivered with| Promise{{🤝 Promises}}
+    Promise --> |resolve to a| Response{📤 Response}
+    Promise --> |join the| EventLoop{{Event Loop 🔁}}
+```
+
+To get data from a server, we make a request with `fetch`. We act on what comes back: the response. But what happens in the middle? We already know that JavaScript is single-threaded: it can only do one thing at a time.
+
+So do we just stop and wait? No! We have a special object to handle this time problem. Run this code in your terminal:
```
illicitonion, line 21: Is this asking them to run this in `node` or in the web inspector console?

content/en/js3/blocks/then/index.md
```diff
@@ -0,0 +1,56 @@
...
+    .then((htmlUrl) => console.log(htmlUrl));
+};
+getProfile("https://api.github.com/users/SallyMcGrath");
+```
+
+So `then` returns a _new_ `Promise`, and you can call `then` again on the new object. You can chain Promises in ever more complex dependent steps. This is called [Promise chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#chaining).
+
+It's important to understand some of what is happening with Promises and `then`. But for the most part, you will not be writing code in this style.
+
+You will be writing ([declaring](/guides/paradigm/)) a description of what you want to happen and a [loosely-coupled](https://en.wikipedia.org/wiki/Loose_coupling) set of stateless, single-purpose functions to make it so. To help you stay thinking in this way, let's explore async/await.
```
illicitonion, line 56: This line feels a bit abstract and not very explained... I think we should either expand it or remove it - which do you prefer?

content/en/js3/blocks/async-await/index.md
```diff
@@ -0,0 +1,69 @@
...
++++
+
+```mermaid
+graph LR
+    Promise{{🤝 Promises}} --> |syntax| async{{🏃‍♂️ async}}
+    async --> |syntax| await{{📭 await}}
+    await --> |resolves to| Response{{📤 Response}}
+```
+
+Async/await is {{<tooltip title="syntactic sugar">}}A simpler, or "sweeter" way to write the same thing. The code works the same under the hood, but it's easier to read. {{</tooltip>}} for Promises. We group them together: async/await, because we {{<tooltip title="use them together. ">}}We can only use `await` inside an `async` function or at the [top level](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) of a module.{{</tooltip>}}
```
illicitonion, line 19: AFAICT it's not possible to focus on the tooltip, so the "top level" link here isn't clickable - when I move the mouse to the tooltip, it gets hidden.

content/en/js3/blocks/async-await/index.md
```diff
@@ -0,0 +1,69 @@
...
+
+We use the [`async`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) [keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords) to define a function that returns a Promise. An async function always returns a Promise.
```
illicitonion, line 21: For some reason the keywords anchor doesn't actually scroll to the right place in this page for me 🤷

content/en/js3/blocks/catch/index.md
```diff
@@ -0,0 +1,44 @@
...
+headless = true
+time = 10
+facilitation = false
+emoji= '🧩'
+[objectives]
+    1='Define a try/catch block'
+    2='Handle an error'
++++
+
+We can handle errors with a try/catch block. We can use the `try` keyword to _try_ to do something, and if it fails, `catch` the {{<tooltip title="error">}}An [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) is a global object produced when something goes wrong. We can throw an [Error](https://www.w3schools.com/js/js_errors.asp) manually with the `throw` keyword.{{</tooltip>}} We can use try/catch in both synchronous and asynchronous code.
```
illicitonion, line 12: Again, tooltip links don't work?

content/en/js3/blocks/fetch-films/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+{{</tab>}}
+{{</tabs>}}
+
+We've added a `try...catch` block to handle any errors that might occur. We've also added `await` to the `fetch` and `response.json()` calls. This means that the function will sensibly wait for the `fetch` to complete before trying to get the json data from the response.
+
+In our last implementation, we called the render function straight away. This time, we need to wait for the films to be fetched before we can render them. Write a new async function to initialise our app. Try to write it yourself first, then check your understanding below.
+
+<details>
+<summary>Your `init` function should look something like this:</summary>
```
illicitonion, line 63: Markdown doesn't render in `<summary>` tags unless they have an empty newline at the start of them, which means the `init` is a literal \`init\`... See DM for more details

illicitonion (commented):
Generally looks good, thanks for putting this together! I opened https://github.com/CodeYourFuture/curriculum/pull/486 with a few edits.

A few of other high-level thoughts to discuss:
1. It feels like we should've introduced `try`/`catch` ´before now. Any thoughts on where we should insert it?
2. I'd quite like us to lean into error handling a bit more - currently most of our trainees think that `console.error` is suitable error handling, but I'd like us to start guiding people towards thinking "How can I show the user that something went wrong, and what they can/should do about it?" - any thoughts as to where/how/when to start doing this?
3. Do we want to talk about errors in _Promises_ at all (i.e. reject and/or talk a bit more about what the `console.log`'d `Promise` values we show are), or are we happy to just say "use `async`/`await` and `try`/`catch`?

content/en/js3/blocks/promises/index.md
SallyMcGrath: Yes you can almost see me realising they haven't done try/catch in the text, can't you? 😂 

I had them running stuff in the browser but then I realised that has a variance we cannot control: CORS settings on the website apply to the console so if they open their console on MDN they will get different results to W3Schools or us. So then I moved it all to the REPL.  Let's specify REPL throughout

content/en/js3/blocks/then/index.md
SallyMcGrath: Yeah, I agree.

Here's my thinking here: at about this point I often get into long discussions with trainees about why we want them to use map, why we keep harassing them to break up their functions, etc. They don't get the point of it and they feel annoyed by it, like we're deliberately making things harder just to vex them. My best response to this has generally been to explain we're learning a style of programming, that they can go read about if they like. And that there are other styles and that's fine, but we're doing this for now. JS3 is IME when these points reach a crescendo, so I'd like to address it. What do you think? How could you rephrase?

content/en/js3/blocks/async-await/index.md
SallyMcGrath: Oh hm, I can fix that -- I habitually tab so I didn't notice. Make a ticket?

content/en/js3/blocks/async-await/index.md
SallyMcGrath: I KNOW IT IS SO ANNOYING

content/en/js3/blocks/async-await/index.md
SallyMcGrath: #490 there you go

content/en/js3/blocks/fetch-films/index.md
SallyMcGrath: #488  / #489  💥

SallyMcGrath:
> Generally looks good, thanks for putting this together! I opened #486 with a few edits.
> 
> A few of other high-level thoughts to discuss:
> 
> 1. It feels like we should've introduced `try`/`catch` ´before now. Any thoughts on where we should insert it?

JS2 - 4 seems the obvious place: https://curriculum.codeyourfuture.io/js2/sprints/4/prep/   IT EMPTY 🕳️ 

> 2. I'd quite like us to lean into error handling a bit more - currently most of our trainees think that `console.error` is suitable error handling, but I'd like us to start guiding people towards thinking "How can I show the user that something went wrong, and what they can/should do about it?" - any thoughts as to where/how/when to start doing this?

So I would say a small piece in JS2 - 4 and then you can have a max 45 minute workshop in JS3 - 4. I had planned this workshop to be web components but I think it's too much and they need the time to do their project. 

> 3. Do we want to talk about errors in _Promises_ at all (i.e. reject and/or talk a bit more about what the `console.log`'d `Promise` values we show are), or are we happy to just say "use `async`/`await` and `try`/`catch`?

I think we should move swiftly on from Promises and let them explore these things when they're a bit more confident reading the code and mentally modelling it. 

My main quibble with what I've written here is I've consistently collapsed the event loop and the task queues. I say join the event loop, or queue to the event loop, when it's more accurate to say join the queue. I did this because there are different queues and I didn't want to get into it, but now I wonder if saying join the event loop isn't even more obscuring. What do you think?

content/en/js3/blocks/then/index.md
illicitonion: I'm inclined to say: Let's remove this from the prep, and try to insert a 30 minute workshop into the day to actively have the discussion (which we can maybe try to put together in the future, through some discussion, rather than trying to make Right Now)?

I think this is something that it's hard to really understand the value of without working through examples (and I'm slightly skeptical that with the current course content we can even motivate that very well - really it's more long-term projects, approaching others' code, and testability where you start seeing the values of this)...

Let's maybe discuss synchronously how these discussions tend to go?

illicitonion:
> > Generally looks good, thanks for putting this together! I opened #486 with a few edits.
> > A few of other high-level thoughts to discuss:
> > 
> > 1. It feels like we should've introduced `try`/`catch` ´before now. Any thoughts on where we should insert it?
> 
> JS2 - 4 seems the obvious place: https://curriculum.codeyourfuture.io/js2/sprints/4/prep/ IT EMPTY 🕳️

Filed https://github.com/CodeYourFuture/curriculum/issues/495

> > 2. I'd quite like us to lean into error handling a bit more - currently most of our trainees think that `console.error` is suitable error handling, but I'd like us to start guiding people towards thinking "How can I show the user that something went wrong, and what they can/should do about it?" - any thoughts as to where/how/when to start doing this?
> 
> So I would say a small piece in JS2 - 4 and then you can have a max 45 minute workshop in JS3 - 4. I had planned this workshop to be web components but I think it's too much and they need the time to do their project.

Sounds good - let's flesh out that discussion on the issue above.

> > 3. Do we want to talk about errors in _Promises_ at all (i.e. reject and/or talk a bit more about what the `console.log`'d `Promise` values we show are), or are we happy to just say "use `async`/`await` and `try`/`catch`?
> 
> I think we should move swiftly on from Promises and let them explore these things when they're a bit more confident reading the code and mentally modelling it.

Sounds good, let's leave this in exactly the level of detail you've gone into - we can see how this goes with the first few classes and work out if there's anything we want to change from there.

> My main quibble with what I've written here is I've consistently collapsed the event loop and the task queues. I say join the event loop, or queue to the event loop, when it's more accurate to say join the queue. I did this because there are different queues and I didn't want to get into it, but now I wonder if saying join the event loop isn't even more obscuring. What do you think?

I think this is a completely fine level of detail for where our trainees are and what they're trying to do. These distinctions are _interesting_ but feel quite internal implementation detailsy of JS engines rather than something someone needs to fully understand in order to be productive or write high-quality code?

illicitonion (approved):
I pushed a commit with the last few edits, including removing the declarative sentence - feel free to either merge or push back :)

Issue #129, opened 2023-06-19
Feature: kitchen sink


SallyMcGrath:
**Which module(s) and week(s) does this change affect?**
Module(s):
Week(s):

**What is the work that needs to be done?**

create a kitchen sink page with all possible components on it:
page header, blocks, youtube, tabs, codemirror, codefence, mermaid, timeline, breadcrumbs, card

**Why is this work important to do?**

useful for testing
very useful for trainees

**Additional context**

**Who might need to know about this change?**

illicitonion:
This effectively got done by https://common.codeyourfuture.io/

PR #493, opened 2024-01-01
add new slide block type

SallyMcGrath:
## What does this change?

Module: all
Week(s): all

## Checklist

## Description

I made a slide block. You can use it to set up slide decks with solo view or just intersperse hero images. At the moment it's just assuming you're linking a single image - though obviously you could expand this.

I made it purely because I want to put some good pictures I found in the React module, but I don't want them on the page with text. I want them solo. I could just make a local content block for each and include this single image, but it felt sufficiently distinguished to become a block.  

In this case I reconstruct the markdown for a figure and send it to the existing render hook for images. 

It expects
```
name="the usual heading, and you could hide this"
src="https://github.com/MaggieAppleton/react-metaphors-slides/blob/master/PNGs/MetaphorsofReact_2.0.017.png"
caption="figcaption and any attr"
```

### DIFF
--- a/layouts/partials/block/data.html
+++ b/layouts/partials/block/data.html
@@ -26,6 +26,7 @@
 {{ .Scratch.SetInMap "blockData" "sot" $src }}
 {{ .Scratch.SetInMap "blockData" "api" $src }}
 {{ .Scratch.SetInMap "blockData" "isShortcode" false }}
+{{ .Scratch.SetInMap "blockData" "caption" .caption | default "" }}
 {{/* Override any block time */}}
 {{ .Scratch.SetInMap "blockData" "time" .time | default "" }}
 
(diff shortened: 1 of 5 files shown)


illicitonion:
Would it be possible to add an example of this? Either by pulling in one of the react things here, or as one piece of https://github.com/CodeYourFuture/curriculum/issues/129?

SallyMcGrath:
Screenshot 2024-01-01 at 21 06 13

SallyMcGrath:
> Would it be possible to add an example of this? Either by pulling in one of the react things here, or as one piece of #129?

https://github.com/CodeYourFuture/curriculum/pull/493/commits/1d9a02a21e2beb55a39a6ca67f8274503d05ba15

makes => https://deploy-preview-493--cyf-curriculum.netlify.app/react/prep/

layouts/partials/block/data.html
```diff
@@ -41,37 +42,44 @@
   ======================
 */}}
 
-{{ if "github" | in $src }}
-  {{/* Change base URL to api.github.com */}}
-  {{ $newSrc := replace $src "https://github.com/CodeYourFuture/" "https://api.github.com/repos/CodeYourFuture/" }}
+{{ if or "png" "webp" "jpg" "jpeg" "gif" | in $src }}
```
illicitonion, line 45: I'm increasingly feeling like we maybe want an explicit "block type" tag in blocks, rather than inferring from what the URL looks like... WDYT?

illicitonion (approved):
Thanks for the preview!

layouts/partials/block/data.html
SallyMcGrath: I think about this from time to time. I just really appreciate the lower friction from having this partial do it for me.

PR #494, opened 2024-01-01
Update README to point users to use dotenv

sztupy:
## What does this change?

Minor change to update the README to point people to use dotenv instead supplying secret values on the command line. The `dotenv` command while non standard is included in the `package.json` so is installed using `npm i`


illicitonion (commented):
LGTM, thanks! I filed https://github.com/sztupy/curriculum/pull/1 against your branch because this didn't quite work for me out of the box, please take a look!

illicitonion (approved):
Thanks!

PR #498, opened 2024-01-02
Error on missing local blocks

illicitonion:
And fix up the three references to blocks which don't exist.


SallyMcGrath (approved):
I think we moved to using a more general study group block and setting a time on it after this module

PR #504, opened 2024-01-03
Add overview page

illicitonion:
It can be hard to tell at a glance what we teach when.

This adds a new page which collates each of the titles of each of the prep blocks.

I have some improvements I'm planning to make, but I think this is a stand-alone useful piece on its own.

### DIFF
--- a/content/en/blocks/backlog/index.md
+++ b/content/en/blocks/backlog/index.md
@@ -3,6 +3,7 @@ title="Backlog"
 headless="true"
 time= 30
 vocabulary="Backlog"
+hide_from_overview=true
 [objectives]
     1="Find the backlog"
     2="Copy your tickets to your own backlog"
--- a/content/en/blocks/spaced-repetition/index.md
+++ b/content/en/blocks/spaced-repetition/index.md
@@ -2,6 +2,7 @@
 title="Spaced Repetition"
 headless="true"
 time= 10
+hide_from_overview=true
 +++
 
 {{<tabs name="Scheduled check-ins">}}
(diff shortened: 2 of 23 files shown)


SallyMcGrath (changes requested):
Nice!

I think there's a simpler way to do this, though. This looks like a taxonomy to me. 

https://gohugo.io/content-management/taxonomies/

You don't have to stick with categories or tags - you can define any taxonomy you like. So you can stick with theme in the front matter, and declare it as a taxonomy in hugo.toml  (and rm hide_from_overview as you won't need that any more) Edit: theme**s** (oh hugo...)

Here's an example on CYF+ which creates versions https://github.com/CodeYourFuture/immersive-go-course/blob/main/website/config.toml

That creates this overview:
https://systems.codeyourfuture.io/versions/

And the version page

https://systems.codeyourfuture.io/versions/1-0-0/

You can then leave a lot of the hard work to Hugo.  Make sense?

illicitonion:
I spent a while trying to make taxonomies work, and ran into issues because it seems hard to group everything together on one page - Hugo seems to somewhat strongly assume you want separate pages for your hierarchies, and it gets fiddlier because the blocks we're making the `<li>`s from aren't themselves pages, so I _think_ we need to do this by manually looping rather than auto-magic at least on the block level, at which point we might as well be looping.

I put together https://github.com/CodeYourFuture/curriculum/pull/530 which I think is a little cleaner, but fundamentally takes the same shape of approach.

Let me know which you prefer, or if there's something big I'm missing!

Issue #503, opened 2024-01-03
Errors Logged to Console


SallyMcGrath:
**Describe the bug**

On https://curriculum.codeyourfuture.io/ 

The console logs these errors

```console
Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
app.min.js:1 Uncaught TypeError: Cannot read properties of null (reading 'querySelector')
    at app.min.js:1:842
```

**What is wrong with this content?**

It's annoying me and that is a Great Crime.

**How could this be fixed?**

Catch the error, handle the null.

**Link to content**

https://curriculum.codeyourfuture.io/

It's in app.js

**Additional context**

**Who else might need to know about this?**

SallyMcGrath:
#507

PR #507, opened 2024-01-04
Fix console error for alert (#503)

metinbaris:
What does this change?
Module: all
Week(s): all

Checklist
 I have read the [contributing guidelines](https://github.com/CodeYourFuture/curriculum/pull/CONTRIBUTING.MD)
 I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
 I have run my code to check it works
 My changes follow our [Style Guide](https://curriculum.codeyourfuture.io/guides/code-style-guide)

Description :

There was a console error regarding element is not defined for alerts on the view in app.js
Moved this to a seperated class and fixed adding event listeners if the element is not loaded.

Note: I couldn't test the alert, so if you know a possible test case for message alerts please let me know :)

Who needs to know about this?
@SallyMcGrath

### DIFF
--- a/assets/scripts/alert-message.js
+++ b/assets/scripts/alert-message.js
@@ -1,0 +1,53 @@
+class AlertMessage {
+  constructor() {
+    const alertEl = document.querySelector(".c-alert");
+    if (alertEl instanceof Element === false) return;
+
+    this.alertEl = alertEl;
+
+    this.attachCloseListener();
+    this.show();
+  }
+
+  attachCloseListener() {
+    const alertClose = this.alertEl.querySelector(".close");
+
+    alertClose.addEventListener("click", (event) => {
+      this.alertEl.setAttribute("hidden", true);
+    });
+  }
+
+  getUrlAlertParameters() {
+    // Extract message and error parameters from the URL
+    const urlParams = new URLSearchParams(window.location.search);
+    const message = urlParams.get("message");
+    const error = urlParams.get("error");
+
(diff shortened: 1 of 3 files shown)


SallyMcGrath:
Please include context in your PR

berkeli:
> (Please include context in your PR.) Can you make a guess as to what is happening with the Performance crash? (Check Lighthouse logging)
> 
> 85 is too low to accept, so would be good to understand what is going on.

I don't think the performance drop has anything to do with this PR tbh, might be a fluke :/ I've done some lighthouse runs locally and getting the same number for the following scenarios:
main branch: 92
this branch: 92
[my branch with a fix](https://github.com/CodeYourFuture/curriculum/compare/bugfix-console-error?expand=1): 92 (wasn't aware of this PR so was putting together a fix)

@metinbaris I'm happy with these changes, but we could optimize it a bit more by conditionally loading the script only when issues are displayed, refer to my branch

metinbaris:
@berkeli, thank you for the review. Instead of placing it inside scripts.js, yes, it makes much more sense!

I initially believed that deferring the script itself should be sufficient, which is why I removed them from the DOMContentLoaded listener of the window.

Thanks also for the test explanation :)

berkeli (approved):

PR #530, opened 2024-01-14
Add overview page

illicitonion:
It can be hard to tell at a glance what we teach when.

This adds a new page which collates each of the titles of each of the prep blocks.

I have some improvements I'm planning to make, but I think this is a stand-alone useful piece on its own.


SallyMcGrath:
I has questions about this. I think I thought this was a more complex multiple-association task.

1. Apart from in fundamentals, it seems like you are actually showing the prep blocks in order per sprint, with a description?
2. You could just update the descriptions on the sprints and then print the local blocks.

Given this, there's somewhere else you're trying to take this. What is the future direction that requires this complexity?

SallyMcGrath (approved):
I'm approving this because I agree this system is better. I've asked why you're not just using the sprint description given the way this actually seems to play out on the page.

However we can stop blocking and iterate on this instead.

PR #546, opened 2024-01-25
Fix environment setup in new structure

sztupy:
The GITHUB TOKEN's environment name has changed recently but this was not reflected in the readme and the .env configs. This change also makes sure that the .env file values are proeprly picked up, and that the README is updated to help new users be able to set up their local dev environment quicker


SallyMcGrath (approved):
LGTM ❤️

PR #515, opened 2024-01-09
Add guides for deployments

sztupy:
## What does this change?

This changes the following:

1. Move all deployment related guides (Previously Netlify and Render) under a "Deployment" section
2. Add some notes on what deployment is, what free options there are and what the limitations of each of these options are
3. Adds section for fly.io deployment
4. Adds section for backend deployment on Netlify
5. Adds section for registering for Supabase

Note: this is WIP, some of the guides would be meant to accompany the re-done Full-Stack-Assessment (and potentially the Final Projects), which is still in works. The links on some pages therefore still link to my own github for examples instead of CYF ones, that would be fixed later. Also not sure whether the added guides adhere to our pedagogical standards - they are mostly how I explained deployments to GLA-6, and try to answer questions that were raised during classes.

## Checklist

- [X] I have read the [contributing guidelines](CONTRIBUTING.MD)
- [X] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- [X] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
- [X] I have run my code to check it works
- [ ] My changes follow our [Style Guide](https://curriculum.codeyourfuture.io/guides/code-style-guide)

PR #555, opened 2024-01-27
Refactor Netlify deployment, adding backend notes

sztupy:
For more info see https://github.com/CodeYourFuture/curriculum/pull/515

This is splitting out that PR into multiple parts. This is part 2 - Netlify


SallyMcGrath:
Amazing stuff.

@sztupy because there's so much to get through, I've pushed changed directly. My changes here are copy edits, not super concerned with content but more form and style. I've simplified the language, added accessibility features like headings, accessible link text, and fixed some of the toml up (there's some legacy yaml in from docusaurus still). 

If you have strong objections, we can talk, but take a look and see what you think

SallyMcGrath (approved):
LGTM!

PR #556, opened 2024-01-27
Refactor Render deployment guide

sztupy:
For more info see https://github.com/CodeYourFuture/curriculum/pull/515

This is splitting out that PR into multiple parts. This is part 3 - Render


SallyMcGrath:
Yeah I think let's move all these guides into these folders:

Contributing
Deploying
Reviewing
Testing

Issue #611, opened 2024-03-04
Feature: tracks and modules view on home page


SallyMcGrath:
**Which module(s) and week(s) does this change affect?**
all

**What is the work that needs to be done?**

Right now our homepage is linear, showing a straight line, but users doing optional modules, tracks, or other programmes (eg DLP, volunteer modules) aren't in this timeline.

Work up some views to express this

**Why is this work important to do?**

See above

**Additional context**

**Who might need to know about this change?**

PR #634, opened 2024-03-18
New home page

SallyMcGrath:
### Common Theme?

New home page

### Org Content?

also touches index of each module, adding to extra menus

## Checklist

## Description

Addresses #611 

In our global syllabus meeting the matter of the home page was raised - two main problems were described:

1. It's not super clear where to start, so I have added a pointing hand and the text "start here" at the first module
2. Tracks, which may begin after JS3 and through which trainees may take different paths, were confusing, so I have grouped them in a way that will make it easier to add lots of tracks and then have trainees pick and mix them to get to the launch

I have styled this lightly as a map - we could go much further with this concept, including an svg path that snakes around the entire curriculum, but tbh that stuff takes ages so I haven't bothered doing it unless we think we need it.

I also haven't done loads of browser testing or anything - let's see if we like the look. Do we like the way we are handling descriptions? Are the columns working? Shall we polish the animations or tone them down?

### DIFF
--- a/common-theme/assets/styles/04-components/card.scss
+++ b/common-theme/assets/styles/04-components/card.scss
@@ -3,6 +3,7 @@
   @include grid-assign(image, description, title);
   background-color: var(--theme-color--block);
   border-radius: var(--theme-border-radius);
+  position: relative;
   grid-template:
     ". .           . .     ." var(--theme-spacing--gutter)
     ". title       . image ." auto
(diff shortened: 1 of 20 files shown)


SallyMcGrath:
This is still open for comment but I'm merging and we can try it for a bit and then comment

Dedekind561 (approved):
Looking cool, let's ship! 🚀

PR #635, opened 2024-03-18
Mermaid fix: prevent loading until intersection

### LINKED ISSUE #411: Mermaid diagrams in CYF workshops (possibly any external markdown) render with an error
**Describe the bug**

If a Mermaid diagram is included in an "external" markdown file, e.g. included via CYF-Workshops then it renders with an error on the curriculum.

See an example with [the debugging workshop](https://curriculum.codeyourfuture.io/js3/sprints/3/day-plan/#break-down) in the day plan for JS3W3 (have to click the "Recap asking questions" accordion):

Screenshot 2023-12-01 at 22 32 19

I'm guessing this console error is related:

```
Uncaught TypeError: Cannot read properties of null (reading 'querySelector')
    at app.min.js:1:1044
```

It borked.

**How could this be fixed?**

Haven't dug into it, but based on the console error, perhaps we're not processing the Mermaid for external markdown files properly?

**Link to content**

https://curriculum.codeyourfuture.io/js3/sprints/3/day-plan/#break-down (have to click the "Recap asking questions" accordion)

**Additional context**

Spotted while reviewing https://github.com/CodeYourFuture/curriculum/pull/400

SallyMcGrath:
1. prevent render
2. initialise on intersection

Fixes #411

An interaction between solo view and mermaid triggered a race mermaid reliably lost. My best guess is that the node the diagram was nested within was removed from the DOM before mermaid had completed its render. Mermaid consequently rendered the SVG with no dimensions, as it sizes itself automagically from its parent dimensions. It was still attaching data-processed="true" though, so it was a blob mess.

I went down a blind alley trying to find out how to re-render mermaid last time I looked at this. I concluded this was a fools' errand this evening and instead wrote an intersection observer that renders mermaid whenever the parent div hoves into view. It doesn't seem to hit performance. I concealed the unsightly FOUT with a small fade on animation.

### DIFF
--- a/common-theme/assets/scripts/mermaid-observer.js
+++ b/common-theme/assets/scripts/mermaid-observer.js
@@ -1,0 +1,32 @@
+import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@latest/dist/mermaid.esm.min.mjs";
+
+mermaid.initialize({ startOnLoad: false });
+// Prevent automatic loading so that mermaid only renders on intersection
+// this is because mermaid regularly loses race conditions
+// and fails to size itself correctly when loaded inside hidden elements that have no dimensions in the DOM
+// like eg tabs, accordions, etc.
+// adapted from https://github.com/mermaid-js/mermaid/issues/311#issuecomment-332557344
+
+const initializeMermaid = (container) => {
+  mermaid.init(undefined, container);
+};
+
+const observeMermaidContainer = () => {
+  const observer = new IntersectionObserver(
+    (entries, observer) => {
+      entries.forEach((entry) => {
+        if (entry.isIntersecting) {
+          initializeMermaid(entry.target);
+          observer.unobserve(entry.target); // Stop observing after initialization
+        }
+      });
+    },
+    { threshold: 0.1 }
+  );
+
+  document.querySelectorAll(".mermaid").forEach((container) => {
+    observer.observe(container);
+  });
+};
+
+document.addEventListener("DOMContentLoaded", observeMermaidContainer);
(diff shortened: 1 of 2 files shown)


Dedekind561:
Never used the intersection observer before! Glad you fixed this 😅 
All working in the deploy preview - full steam ahead... 🚂

Dedekind561 (approved):
As I said before, never used Intersection Observer API - I'm just checking the deploy preview at this point and it looks good to me!

SallyMcGrath:
> As I said before, never used Intersection Observer API - I'm just checking the deploy preview at this point and it looks good to me!

It's mainly used to animate in cards on scroll, but in this case what I did was to prevent rendering until the div was intersecting the viewport. This meant Mermaid always had the correct dimensions when rendering, and it doesn't need to know anything about WHY the div isn't rendering, so it handles all the hidden cases (tabs, solo view, details, etc)

40thieves:
> the node the diagram was nested within was removed from the DOM before mermaid had completed its render

Out of curiosity do you know what is removing this DOM node? Is it mermaid itself?

40thieves:
Also nice fix! 🎉

SallyMcGrath:
> > the node the diagram was nested within was removed from the DOM before mermaid had completed its render
> 
> Out of curiosity do you know what is removing this DOM node? Is it mermaid itself?

Any parent element with the attribute hidden=true is [not rendered.](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#hidden) . We definitely want this behaviour as it's how we manage focus and position for non-sighted contexts. Details elements do this natively. Accessible tabs do this, and our solo view also does this. Our solo view is both reasonably performant and a parent node, so it loads and executes before mermaid can complete. 

So because the parent .mermaid div is literally not rendered but still present, a blob ensues. This is my guess anyway.

40thieves:
> Accessible tabs do this, and our solo view also does this

Ah I see, I hadn't connected this only affected those views. In that case it totally makes sense.

PR #636, opened 2024-03-20
603 Images are broken on https://curriculum.codeyourfuture.io/guides/contributing/supporters/

### LINKED ISSUE #603: Images are broken on https://curriculum.codeyourfuture.io/guides/contributing/supporters/
**Describe the bug**

Images are broken on https://curriculum.codeyourfuture.io/guides/contributing/supporters/

Images are broken

**How could this be fixed?**

Likely to do with the warning  `WARN  The []interface {} data from 'funding.json' overridden by higher precedence []interface {} data already in the data tree`

If so, fix by reversing the merge order on the module mount

**Link to content**

**Additional context**

SallyMcGrath:
Fixes #603

On production, absURL was defaulting to / instead of the site root.

Seems like maybe it's because curriculum is a subdomain? In any case, I have added the url explicitly using baseURL

https://gohugo.io/methods/site/baseurl/

If this is the right fix I have some updates to the funding page, which I will open as a second tiny PR

https://deploy-preview-636--cyf-curriculum.netlify.app/guides/contributing/supporters/

#637

### DIFF
--- a/common-theme/layouts/_default/_markup/render-codeblock-mermaid.html.html
+++ b/common-theme/layouts/_default/_markup/render-codeblock-mermaid.html.html
@@ -13,4 +13,4 @@
 </style>
 <div class="mermaid">{{- .Inner | safeHTML }}</div>
 {{ $mermaid := resources.Get "scripts/mermaid-observer.js" | resources.Minify }}
-<script type="module" src="{{ $mermaid.Permalink }}" defer></script>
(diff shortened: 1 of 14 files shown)


SallyMcGrath:
This is the right fix but it will create phantom bugs on dispoable branch previews because it removes relative urls and therefore triggers weird CORS errors and also will direct you to the production site when you click a link, and so testers will not notice this and think there are no changes

PR #659, opened 2024-03-30
Content: DB module project and workshops WAITING FOR WORKSHOP MERGE

SallyMcGrath:
## What does this change?

This branch has stubbed module project, in preparation for the full stack assessment updates

I will also be linking these workshops when they are merged

week 1  is in: bug reports https://github.com/CodeYourFuture/CYF-Workshops/tree/main/reporting-bugs
week 2 https://github.com/CodeYourFuture/CYF-Workshops/pull/188
week 3 https://github.com/CodeYourFuture/CYF-Workshops/pull/189
week 4 is in - dom merge conflicts

Issue number: #issue-number

### Org Content?

Databases | 1234 | Dayplans, Product| external links to CYF workshops and Full Stack Assessment

## Checklist


sztupy:
Based on the current FSA proposal the trainees would need to know:

**Week 1:**

mandatory:

```
CREATE TABLE - ID and strings
INSERT INTO
SELECT * FROM
```

**Week 2:**

mandatory:

```
INSERT INTO
DELETE FROM
```

**Week 3:** 

mandatory:

```
UPDATE
```

optionally:

```
ALTER TABLE
```

**Week 4:**
optionally:

```
ALTER TABLE 
SELECT - ORDER BY
```

Checking the old syllabus (https://syllabus.codeyourfuture.io/db/week-1/lesson ) we have:

**Week 1:** mostly just SELECT although with lots of filtering. Nothing about CREATE TABLE and INSERT INTO (apart from asking them to run a pre-created SQL). Also no notes on how to intergrate with Node which is started in Week 2 but mostly during Week 3. These are somewhat pre-filled in the assesment however so it's possible that this would not be a major problem. (And the required SQL is on the very simple side)

**Week 2:** we teach INSERT, UPDATE and DELETE, so the week's requirement is covered from the SQL side, but we only teach Node.JS parametrized queries in Week 3 which is something needed here. Again, some slight help might be needed from the FSA exercise, OR we need to move some bits around (like UPDATE to week 3, but some node related stuff to week 2)

Do note that the SQL requirement in the FSA is on the simple side (one table, no joins, no real filters, search only by ID, ordering is only needed as an extra, etc) so if we expand the exercise definitions we might actually be covered, but it's possible we want to move some bits around so at least we mention things to the trainees

So there's definitely some misalignment here

Dedekind561:
If we give the trainees the `initdb.sql` file with some of the additional `CREATE TABLE` and `INSERT INTO` statements then I don't think we'll be that misaligned. From the instructions, trainees will be able to work out most of the stuff. I've captured this issue, which I'll prioritise: https://github.com/CodeYourFuture/curriculum/issues/669

SallyMcGrath:
We also have the extra lessons they do in North West, which we could include if we can figure out how to squeeze it in

https://github.com/KeithBremer/CYF-DB-Module/blob/master/extra/extra-1/extra_sql.md
https://github.com/KeithBremer/CYF-DB-Module/blob/master/extra/extra-2/extra-node.md

KeithBremer:
I'm not sure how the week plans (from sztupy) relate to the FSA plans but from a learning perspective they fall into a classic trap in the development of learning materials: don't _**teach**_ in the order you _**do**_.  I've said before that asking a novice to design tables before they've used them is like asking someone to design a car before they've ever driven one.  The original course materials are deliberately sequenced in the way they are to avoid this.  Teach basic and progressively more complicated SELECT commands (including joins) first, then the learner understands what RDBMS is about.  Only then can they be expected to define a table with any success.
I know that in real life they have to define the tables first before they populate them and start to query/update/delete but that necessity should not govern the order in which they learn.

Just my ha'porth.

sztupy:
@KeithBremer the Full Stack Assesment is done the same time as the DB course, so if we ask them to do something in the FSA that we haven't taught them in the DB course that would probably be problematic from a pedagogical standpoint. And the FSA is just the opposite - you start with some kind of table / data model.

(This wasn't a problem in the old FSA, as Week1&2 did not have any database activity at all - you had to do everything using static files and in-memory arrays. SQL was only required during Week 3, by which time you had already finished the DB module as it was only 3 weeks long)

PR #696, opened 2024-04-14
Add checks for sprint consistency

illicitonion:
We previously had sprints with missing themes.

This will cause a CI failure if we regress this.

I can imagine many other checks we may want to automate in the future :)


SallyMcGrath:
Couple questions:

Is this restricted to org-cyf repo? I can see you've skipped a hard coded list of as-yet-unthemed modules but I can't see a restriction to org-cyf. My feeling is that it's probably ok for us to mandate themes in our repo but it's an annoying constraint for others. 

Secondly at some point we should revise the overview page so it doesn't create output unless it's opted in to, to match the rest of the content creation.

SallyMcGrath (changes requested):
Please respond to my questions from April TY! PS no judgment!

illicitonion:
> Is this restricted to org-cyf repo? I can see you've skipped a hard coded list of as-yet-unthemed modules but I can't see a restriction to org-cyf. My feeling is that it's probably ok for us to mandate themes in our repo but it's an annoying constraint for others.

I've added an `--exclude-roots` flag which lets us exclude specific top-level dirs (currently being used to exclude `org-cyf-launch`), and could extend it to be a pattern if we'd prefer, but I _think_ this is obsolete now that we only host common and org-cyf*?

> Secondly at some point we should revise the overview page so it doesn't create output unless it's opted in to, to match the rest of the content creation.

Sure! How would you like that check to work? Filtering on a sprint being in a `module` menu_level?

SallyMcGrath:
> > Is this restricted to org-cyf repo? I can see you've skipped a hard coded list of as-yet-unthemed modules but I can't see a restriction to org-cyf. My feeling is that it's probably ok for us to mandate themes in our repo but it's an annoying constraint for others.
> 
> I've added an `--exclude-roots` flag which lets us exclude specific top-level dirs (currently being used to exclude `org-cyf-launch`), and could extend it to be a pattern if we'd prefer, but I _think_ this is obsolete now that we only host common and org-cyf*?


Yes now I've migrated migracode to https://github.com/Migracode-Barcelona/curriculum this objection is moot. 

> > Secondly at some point we should revise the overview page so it doesn't create output unless it's opted in to, to match the rest of the content creation.
> 
> Sure! How would you like that check to work? Filtering on a sprint being in a `module` menu_level?

I fixed this myself at some point already... #781

SallyMcGrath (approved):

PR #714, opened 2024-05-01
Feature/  Block links

SallyMcGrath:
## What does this change?

Set up a block link partial - this is a block that is just a link. It doesn't do anything fancy, like grab open graph data from the link - I considered this but I don't think it's worth the bytes. Similarly, I will just put some tasty bg images (from cyf+ site) for now. 

In another PR I will then move on to fixing the image bug in this site and let a picture on the block.  This is a small part of the work to por CYF+ to a skinned version of this site. (I've done the skin and now to extend the theme with the parts we want)

### Common Theme?

block/link
shortcodes/blocklink

test with

```toml
[[blocks]]
name="Test block link"
src="https://react.dev/learn/tutorial-tic-tac-toe"
caption="This is a test block link"
```

or in shortcode

```
{{< blocklink src="https://react.dev/learn/tutorial-tic-tac-toe" name="Block link" caption="Tic tac toe" >}}

{{< blocklink src="https://react.dev/learn/tutorial-tic-tac-toe" name="You have built components that are functions" caption="Tic tac toe" >}}

{{< blocklink src="https://react.dev/learn/tutorial-tic-tac-toe" name="Block link" caption="You have built components that are functions. You have passed data into those functions. You have defined state objects. You have mapped over data and returned new arrays of components. You’ve practically learned React already! It’s like we planned this…" >}}
```
Screenshot 2024-05-01 at 22 18 03
Screenshot 2024-05-01 at 22 18 11


common-theme/layouts/partials/block/data.html
```diff
@@ -83,3 +84,5 @@
   {{ end }}
 
 {{ end }}
+{{/* {{ printf "DEBUG: Scratch contents %v" (.Page.Scratch.Get "blockData") }}
```
illicitonion, line 87: Remove?

common-theme/layouts/partials/block/link.html
```diff
@@ -0,0 +1,20 @@
+{{ $blockData := .Page.Scratch.Get "blockData" }}
+{{ $randomImage := (index (resources.Match "custom-images/backgrounds/**.**" | shuffle | first 1) 0) }}
+{{ $randomImage = $randomImage.Resize "480x" }}
+
+
+<section class="c-block c-block--{{ $blockData.type }}">
+  <a
+    id="{{ $blockData.name |urlize }}"
+    href="{{ $blockData.sot }}"
```
illicitonion, line 9: What does `sot` stand for? Should this be `src`?

illicitonion (approved):
Thank you!!

common-theme/layouts/partials/block/link.html
SallyMcGrath: source of truth - sometimes you need the link and then a transformed link to say an api

{{/* Set the source of truth and api defaults to src */}}
{{ .Scratch.SetInMap "blockData" "sot" $src }}
{{ .Scratch.SetInMap "blockData" "api" $src }}

SallyMcGrath: Oh interesting - I replied via the VSCode integration I just set up (it's what the trainees use) and it didn't add to the thread, just added a new comment.

Replying directly in VSCode seems like something we should figure out because it's something volunteers identify repeatedly about the problems with code review (getting demotivated because trainees ignore them)

PR #719, opened 2024-05-06
Feature: absolute image paths, get remote images

SallyMcGrath:
### Common Theme?

1. _markup/render-image.html

revise render-image hook to do the following:

- update github links to github raw links (instead of jsdelivr)
- get all remote images and process as local instead of hotlinking
- incidentally, perform a mediatype check instead of file type check so we actually know it's a rasterisable image and not just some other file masquerading as same

2. strings/absolute-image-paths.html

then we made a short partial to run over decoded markdown blobs and regex with the path of the repo it's living in NOTE I DID NOT CONSTRUCT THIS REGEX please review my terrible regex 

3. blocks/readme.html

in the readme block, we added a second and final processing step to turn the relative links absolute. It has to be the last step so we can markdownify it inside the assignment. If we don't do this, I think we have to pass the text to scratch to get it back into the readme block context. However, having it as a separate partial is probably best so we can reuse it in other places if we need to like in issues and pd blocks if that comes up.

So it goes 

README => getJSON readme blob and decode
STRINGS/ABS... => regex over decoded text to replace relative paths with the html path
README => then send that to markdownify

inside markdownify
RENDER_IMAGE => replace github html path with github raw location
then getRemote and download that image

then process, resize etc as any other image

#222 

## Checklist

### DIFF
--- a/common-theme/layouts/partials/block/link.html
+++ b/common-theme/layouts/partials/block/link.html
@@ -1,6 +1,8 @@
 {{ $blockData := .Page.Scratch.Get "blockData" }}
 {{ $randomImage := (index (resources.Match "custom-images/backgrounds/**.**" | shuffle | first 1) 0) }}
-{{ $randomImage = $randomImage.Resize "480x" }}
+{{ if $randomImage }}
+  {{ $randomImage = $randomImage.Resize "480x" }}
+{{ end }}
 
 
 <section class="c-block c-block--{{ $blockData.type }}">
--- a/common-theme/layouts/_default/_markup/render-image.html
+++ b/common-theme/layouts/_default/_markup/render-image.html
@@ -40,8 +40,12 @@
 
   {{/* Handle external images directly */}}
   {{- if $isExternal -}}
-    {{ $image = (dict "RelPermalink" $src "Width" 0 "Height" 0) }}
-  {{- end -}}
+    {{ with resources.GetRemote $src }}
+      {{ $image = . }}
+    {{ else }}
+      {{ errorf "Unable to get remote resource %q" $src }}
+    {{ end }}
+  {{ end }}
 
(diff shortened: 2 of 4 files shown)


common-theme/layouts/partials/block/link.html
```diff
@@ -1,6 +1,8 @@
 {{ $blockData := .Page.Scratch.Get "blockData" }}
 {{ $randomImage := (index (resources.Match "custom-images/backgrounds/**.**" | shuffle | first 1) 0) }}
-{{ $randomImage = $randomImage.Resize "480x" }}
+{{ if $randomImage }}
```
SallyMcGrath, line 3: This is just a little guard against an annoying Hugo server fault that was getting in the way of this fix - it is basically irrelevant otherwise

SallyMcGrath:
# Example of current, final problem

For some reason my partial is still trying to transform non images. Here's the final case where an image is not an image

https://github.com/CodeYourFuture/JS3-Module-Project/issues/9

This is shown on https://curriculum.codeyourfuture.io/js3/product/backlog/backlog-4/ 

It is coming through the issues.html in this case

Choose edit  to find the uploaded src (drag and drop) as  `https://github.com/CodeYourFuture/JS3-Module-Project/assets/42203406/fc994768-23ba-454a-8a1f-8b3375e264f8` NOTE THE LACK OF filetype

Screenshot of a website with a drop-down list with the show "Breaking Bad" selected

VIew image to find the actual src, in this case

https://private-user-images.githubusercontent.com/42203406/307878913-fc994768-23ba-454a-8a1f-8b3375e264f8.png

But you need a JWT to view

SallyMcGrath:
YESSSSSSS I FIXED ITTTTTTTT

Dedekind561 (approved):
Tally ho 🚀

PR #743, opened 2024-06-06
FEATURE: stub piscine site

SallyMcGrath:
This is a new tenant for the CYF Piscine site, which should be a standalone site that can be mounted into other sites.

I made this by copying an existing site and junking all the extraneous files.

I think I need to merge this and then I can init and mount this as a module to other sites eg itp
Screenshot 2024-06-06 at 14 35 45


Dedekind561 (approved):
Tally ho...

PR #751, opened 2024-06-17
import and mount guides in org-cyf

Dedekind561:
## What does this change?

- Imports and mounts `guides` into `org-cyf`

## To check

- `cd org-cyf`
- Run `npm run start:dev`
- Double-check that it builds!


SallyMcGrath:
I think this is obsoleted by merging #746 and #747

PR #769, opened 2024-06-30
Two small tests for content

SallyMcGrath:
### Common Theme?

- [x] Yes

Here are two small tests to make sure
1. the success page is compiling learning objectives, as this has broken silently before
2. the day plan has blocks on it, because without blocks it has no value

I have **not** added a third, more controversial test, on line 27 of prep.html

```
{{ else}}
  {{ errorf "%s has no blocks in the front matter. Add blocks or remove the page" $.Page.RelPermalink }}
{{ end }}
```
As the sites would currently fail to build (13 blank views in the SDC). But pls consider this. 

## Checklist

### DIFF
--- a/common-theme/layouts/_default/day-plan.html
+++ b/common-theme/layouts/_default/day-plan.html
@@ -4,7 +4,12 @@
       "c-page-header--toc"
     }}
     {{ partial "page-header.html" . }}
-    <section class="l-page__main c-copy">{{ .Content }}</section>
+    {{ with .Content }}
+      <section class="l-page__main c-copy">{{ . }}</section>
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Amazing, thanks!

I would love to do this for prep too, but agree we should fix up the missing prep views before doing so. In the past with this kind of clean-up I've added an allow-list of known-broken pages, to avoid new breakages coming in, but probably not worth it here :)

SallyMcGrath:
Yeah I feel like with 13 broken pages maybe let's just clean them up and then put the test on - a ticket maybe?

PR #773, opened 2024-07-04
Feature: make issues and github stuff work outside of CYF

SallyMcGrath:
## What does this change?

We had limited this to CYF's org because it felt like dangerous madness not to. But actually MCB need their own repos so I've removed the limit. 

This reminded me of the problem with paginated issues when trainees start PRing to coursework repos -- the coursework gets pushed off page 1. 

So I think I've paginated now? I need to test a bit more, hence draft PR

### Common Theme?

- [x] Yes

This changes the way issues are pulled from github for everyone, which is why this PR is draft until I've tested a bit more.

it affects mainly 

issues.html
fetch-paginated-issues.html <= new partial called recursively

The other changes are config changes to remove the old settings.

Note: this won't build! that's because MCB need to create all these repos or remove the links to them. 

That's happening on this ticket https://github.com/Migracode-Barcelona/Table-of-Contents/issues/1

On my local copy of MCB, I have just deleted all content except JS1, as I have created that test module to build this stuff. 

### Org Content?

I also made a logo just because it was bugging me a lot to look at

## Checklist

- [ ] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)


common-theme/layouts/partials/issues.html
```diff
@@ -1,20 +1,24 @@
 {{ $repo := .Params.backlog }}
-{{ $issuesUrl := print .Site.Params.issuesorgapi $repo "/issues?per_page=100&state=open&direction=asc&type=issue" }}
```
SallyMcGrath, line 2: type issue does nothing to filter out prs so I removed it

common-theme/layouts/partials/issues.html
```diff
@@ -1,20 +1,24 @@
...
-<!-- api call -->
+{{/* <!-- api call --> */}}
 {{ $headers := (dict) }}
 {{ if ne (os.Getenv "HUGO_CURRICULUM_GITHUB_BEARER_TOKEN") "" }}
   {{ $headers = merge $headers (dict "Authorization" (printf "Bearer %s" (getenv "HUGO_CURRICULUM_GITHUB_BEARER_TOKEN"))) }}
 {{ end }}
-{{ $issues := getJSON $issuesUrl $headers }}
-<!-- if no object show error -->
+{{/* <!-- walk through paginated issues --> */}}
+{{ .Page.Scratch.Set "issues" slice }}
```
SallyMcGrath, line 11: same deal as block/block.html => block/data.html

org-mcb/config/issues-are-cached-and-incomplete/config.toml
```diff
@@ -0,0 +1,12 @@
+# This config is optimised for speed of local interation of general Hugo set-up without worrying about issue freshness or completeness.
+# To build with this config, run `hugo server --environment issues-are-cached-and-incomplete`.
+
+[params]
+# Don't use the paginating proxy, because it's slow.
+# This means some issues will be missing when fetched.
+issuesorgapi = "https://api.github.com/repos/Migracode-Barcelona/"
```
SallyMcGrath, line 7: we can probably rm all this, though I did like just skipping pagination on local

org-mcb/hugo.toml
```diff
@@ -16,13 +16,11 @@ description = "A free and open source software development programme."
 main_website = "https://migracode.org/"
 main_org_name = "MigraCode Barcelona"
 our_name="MCB"
-org = "https://github.com/CodeYourFuture"
-repo = "https://github.com/CodeYourFuture/curriculum/"
+org = "https://github.com/Migracode-Barcelona"
```
SallyMcGrath, line 19: this all Just Works 🎉

SallyMcGrath:
I gott a test this a lot more, but of course am getting rate limited. Another strong argument for no crawling on dev!

common-theme/layouts/partials/fetch-paginated-issues.html
```diff
@@ -0,0 +1,22 @@
...
+{{/* get the issues we have so far in our Scratch */}}
+{{ $fetchedIssues := .Scratch.Get "issues" }}
+
+{{ if and $pageOfIssues (gt (len $pageOfIssues) 0) }}
+  {{ range $pageOfIssues }}
+    {{ $fetchedIssues = $fetchedIssues | append . }}
+  {{ end }}
+  {{ .Scratch.Set "issues" $fetchedIssues }}
+  {{/* If we got 100 issues, there might be more. Let's fetch the next page */}}
+  {{ if eq (len $pageOfIssues) 100 }}
```
illicitonion, line 18: I'm not sure this is correct, I think GitHub will return non-100-item pages but still need pagination - we could maybe just always try to get page+1 and see if it has results?

The way my little proxy does it is to look in the `link` header for a `next` link, I'm not sure if there's a handy way to get at headers from a hugo response, but I would assume not?

illicitonion (commented):
Generally LGTM except I'm not sure the pagination check is a reliable one - I'm pretty sure I saw partial responses still have nexts (probably because e.g. GitHub treats PRs as issues and maybe filters after it paginates or something)

common-theme/layouts/partials/fetch-paginated-issues.html
SallyMcGrath: Hrm, it looks like there's a new way https://gohugo.io/functions/resources/getremote/#remote-data in the latest version and I can maybe get into them that way. Will look

SallyMcGrath: Coming back to say - it looks like I can't get full details yet https://github.com/gohugoio/hugo/issues/12521 

When this comes out I can do it that way and I will. In any case I will update to the new way of getting json in advance.

If a request is set to get 100 and there are fewer than 100 remaining, in what scenario will GitHub paginate? (Genuine question -- help me think through this so I can sort this for MCB)

SallyMcGrath: And yy re your point above: GH does treat PRs as issues, so that response contains both and we filter afterwards

illicitonion: > If a request is set to get 100 and there are fewer than 100 remaining, in what scenario will GitHub paginate?

I'm not sure, but I'm pretty sure I've observed it. If you can convince yourself it doesn't ever do this, I won't argue :)

SallyMcGrath: I'm not convinced it won't happen, I just don't know _how_ it would happen. 

Also I think we might have to start versioning this platform... 🤮

SallyMcGrath: I'm going to have to push ahead with this because MCB are about to launch

PR #774, opened 2024-07-04
sdc: Add draft topics, blocks, and learning objectives

illicitonion:


40thieves:
Thanks for getting this started! 🎉

Some rough initial thoughts from me:

I do think it's good to have something of a "hook" to get trainees pulled onto the course, and I'm not sure binary numbers is a great hook? But perhaps that's just my experience (I've never got on well with binary). And it is balanced with some CLI tools.

Beyond my point above, it feels like there's decent "pacing" of "hands-on" vs "abstract", starting relatively hands-on, moving to quite abstract, then back to hands-on before finishing out fairly abstract. I must admit I'm somewhat surprised to see the Logic topic in the middle - I would expected you of all people to have put that first :)

I'm not sure I agree with the naming of the Legacy topic - in the LOs you've framed it more as "existing code", which admittedly isn't as catchy but I think is more what we should go after. I think we should avoid the implication that any existing code is "legacy" as it's such a loaded term in industry.

SallyMcGrath:
> Thanks for getting this started! 🎉
> 
> Some rough initial thoughts from me:
> 
> I do think it's good to have something of a "hook" to get trainees pulled onto the course, and I'm not sure binary numbers is a great hook? But perhaps that's just my experience (I've never got on well with binary). And it is balanced with some CLI tools.
> 
> Beyond my point above, it feels like there's decent "pacing" of "hands-on" vs "abstract", starting relatively hands-on, moving to quite abstract, then back to hands-on before finishing out fairly abstract. I must admit I'm somewhat surprised to see the Logic topic in the middle - I would expected you of all people to have put that first :)
> 
> I'm not sure I agree with the naming of the Legacy topic - in the LOs you've framed it more as "existing code", which admittedly isn't as catchy but I think is more what we should go after. I think we should avoid the implication that any existing code is "legacy" as it's such a loaded term in industry.

What about "Other people's code" ?

SallyMcGrath:
BTW I'd say merge this and we can get revising after Piscine -- it's good to have something drafty up that ITP and Piscine can look over easily.

Maybe just put a note on saying DRAFT

SallyMcGrath (approved):
merge as draft

PR #776, opened 2024-07-04
Add netlify deploy script

illicitonion:
We were previously not setting --baseURL which means that deploy previews have internal links to the prod deploys rather than the deploy previews.

Instead we introduce a script so that we can edit for all of the sites without needing to go change several site configs each time we want to make a change.

### DIFF
--- a/common-content/deploy-netlify.sh
+++ b/common-content/deploy-netlify.sh
@@ -1,0 +1,1 @@
+../deploy-netlify.sh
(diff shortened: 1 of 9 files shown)


SallyMcGrath:
Does each thing just run the shared netlify deploy shell script? Or am I missing smth?

illicitonion:
Yeah, each runs the same script, but they run from different working directories which means they all do the right thing in context.

SallyMcGrath (approved):

PR #817, opened 2024-07-18
Feature: remove migracode and remove limit to CYF org

SallyMcGrath:
MCB curriculum has moved to https://github.com/Migracode-Barcelona/curriculum

This PR does the following

1. removes the limit to the CYF org on Github
2. removes Daniel's proxy for paginated issues in favour of a recursive partial NOTE this may need more work but what can we do with the time constraints
3. updates all the GetJSON calls to the new GetRemote method
4. extracts the github auth headers to partial that returns the correctly formed map, which is slightly slightly different to the old map
5. improves overall the null checks to use a with pattern and warnings 
6. uses the new .Err on GetRemote (previously this was roll your own)
7. removes the mcb site entirely

I merged a version of this last night and completely **** the bed. This was because

1.  the auth was malformed
2.  valid responses of empty arrays cannot be transformed with unmarshal and cannot be detected with .Err or reflect or len etc, so we have to really specify
`{{ if eq . "[]" }}` exactly. This could be a valid response on issues and pr endpoints, but not on readme or contributors.

I kept this branch and will fix up the conflicts in here instead of trying to rerun the previous PR


SallyMcGrath:
I'm gonna have to try this  even with this rate limiting...

PR #821, opened 2024-07-20
Feature: return proxy, template revision and a sample config

SallyMcGrath:
this removes the recursive partial call in issues.html
then it updates the path of orgapi to Daniel's proxy
it doesn't say orgissuesapi  - we can override all gh api calls that return a list


org-cyf-itp/hugo.toml
```diff
@@ -47,7 +47,7 @@ org = "https://github.com/CodeYourFuture"
 repo = "https://github.com/CodeYourFuture/curriculum/"
 pdrepo = "CYF-PD"
 root = "curriculum"
-orgapi = "https://api.github.com/repos/CodeYourFuture/"
+orgapi = "https://github-issue-proxy.illicitonion.com/repos/CodeYourFuture/"
```
illicitonion, line 50:
```suggestion
orgapi = "https://github-issue-proxy.illicitonion.com/cached/120/repos/CodeYourFuture/" 
```

There are a couple of stray settings in `issuesorgapi` in org-cyf-sdc/hugo.toml and org-cyf-guides/hugo.toml which I think are now unused?

common-theme/layouts/partials/issues.html
```diff
@@ -1,69 +1,76 @@
...
-  <!-- range over issues list and pull out useful data -->
-  {{ range $issues }}
+{{ with resources.GetRemote $issuesUrl $headers }}
+  {{ with .Err }}
+    {{ errorf "Failed %s on %s. Error: %s" $issuesUrl .Page.Title . }}
+  {{ end }}
+
+  {{ with .Content }}
+    {{ if eq . "[]" }}
+      {{ warnf "No issues found at %s" $issuesUrl }}
```
illicitonion, line 15: Maybe error rather than warn? And I'd be tempted to have this as an accumulated check after the loop (i.e. apply the filter in the loop, if we show an issue set something to true, and error if we didn't show at least one issue) - "You set a filter and it ignored all issues" feels like as much of an error as "Your original query without filters had not results"?

illicitonion (approved):
Sure, what could go wrong? :D

illicitonion:
Weird, why is netlify only previewing one of the sites?

org-cyf-itp/hugo.toml
SallyMcGrath: Yeah I'll run over the rest now and update

common-theme/layouts/partials/issues.html
SallyMcGrath: So, the question here is how much do we want builds to fail, basically, based on external factors.

If there are no issues set on a repo we aren't personally in charge of, like a course we are not running, should that fail our build? When PD can't merge their blocks because there's a no response error they cannot and do not fix it. Instead I have to run around and fix it and a lot of time is wasted on this sort of thing that I don't necessarily have. 

I don't disagree that it would be more rigorous, it's just that there's only me here.

SallyMcGrath:
> Weird, why is netlify only previewing one of the sites?

I turned em off as I was working through reducing requests. Noted it in syllabus #

PR #843, opened 2024-07-31
Feedback suggestions on Piscine skeleton

illicitonion:


SallyMcGrath (approved):
yes I want all of these, thank you!

Issue #469, opened 2023-12-20
Document the architecture of the curriculum


40thieves:
**Which module(s) and week(s) does this change affect?**
Module(s): N/A
Week(s): N/A

**What is the work that needs to be done?**

Add some documentation of how the curriculum works at a high level.

I think this is different from #129, which documents specific components available within the content. The goal here is how the whole thing hangs together.

**Why is this work important to do?**

It's helpful for newcomers to have some understanding, especially when fixing bugs. This is especially important in light of our partnership with Migracode as we'll hopefully have a lot of new folks looking to contribute!

**Who might need to know about this change?**

I'm guessing this is a job for @SallyMcGrath!

PR #844, opened 2024-08-02
Feature: block viewer  for docs 

SallyMcGrath:
Addresses #469 

I've written another documentation site. I threw away the first one because it didn't address the real questions users have.

To make this work I need to do three main bits of work:

1. Build a new view called block-viewer. I've put this in common-theme because people might want it for something else later. 

This view takes any blocks stored in a folder and displays them in a solo view format so you can page through them to take a quick look. 

Then the rest of the tweaks are stuff I encountered as I wrote the docs today, and are listed in the commit messages.

Once this PR is merged, move on to the actual docs site, which I will open soon common-docs #846 

Then the final piece is making some changes to common-content. Where previously we used headless=true to prevent render, this also suppresses section listing. (Even though it claims not to, it does.) #845 

So instead we're updating it to the new system, build cascade and turning off rendering https://gohugo.io/content-management/build-options

I will add an index to each module folder with a description and so on, and that will create a navigable directory of blocks to view.

## To review

Honestly the main goal here is to check we haven't broken anything. Obviously I have run over things myself but I am only a human bean.

This PR mainly messes around with

- page header -- shows on every page, allows us to manually force a table of contents by passing a slice of headings to the Page Scratch, along with our other manual TOC creation
- time -- I revised this slightly to a) simplify it and b) make it not break if we don't have any $blockData for whatever reason. This should also mean you can use timers on non-block views like single.
- data - I revised the PD block to also work with mounted content as I confidently explained it did this in the docs and then realised the check was missing

### DIFF
--- a/common-theme/archetypes/blocks/index.md
+++ b/common-theme/archetypes/blocks/index.md
@@ -1,7 +1,11 @@
 +++
 title = '{{ replace .Name "-" " " | title }}'
 time = 30
-headless = true
+[[cascade]]
+  [cascade.build]
+    list = 'local'
+    publishResources = true
+    render = 'never'
 facilitation = false
 hide_from_overview=false
 threads = ['unassigned']
--- a/common-theme/assets/styles/04-components/page-header.scss
+++ b/common-theme/assets/styles/04-components/page-header.scss
@@ -79,6 +79,12 @@
       background-color: transparent;
       text-shadow: none;
     }
+    .c-page-header--solo & {
+      max-width: 8ch;
+      overflow-x: hidden;
+      white-space: nowrap;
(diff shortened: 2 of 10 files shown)


illicitonion (approved):
A bunch of blocks seem to now have 0-duration times. Examples:

https://deploy-preview-844--cyf-curriculum.netlify.app/js3/sprints/1/day-plan/
Energiser, Teamwork Project S1, Morning Break, Community Lunch, Components are functions workshop all are 0

https://deploy-preview-844--cyf-curriculum.netlify.app/js3/product/prep/ went from 60 minutes to 0 minutes

Other than that, LGTM in terms of not breaking things - I'll review how it works in #846

PR #845, opened 2024-08-02
Feature/ Content: Common Content mass front matter change

SallyMcGrath:
## What does this change?

1. Change all instances of `headless=true` in blocks front matter to

```toml
[build]
    render = false
    list = 'local'
    publishResources = false
```

https://gohugo.io/content-management/build-options/

2. Add an index file to each folder, so these can show on common-docs block viewer

### Common Content?

Yes ALL blocks in common content

This is part 2 of 3 linked PRS

#844 <== common-theme changes
#845 <== common-content changes
#846 <== common-docs site created


illicitonion (approved):
It feels like there are two useful tests we could automate here:
1. Pages which should have the `[build]` section you've inserted have it
2. ?Every directory has an index of some description

Once I address https://github.com/CodeYourFuture/curriculum/pull/696 and get it merged, I'd be very happy to either write these checks, or pair with you to do so; can we try to enumerate the conditions where each is expected/required?

SallyMcGrath:
> It feels like there are two useful tests we could automate here:
> 
> 1. Pages which should have the `[build]` section you've inserted have it
> 2. ?Every directory has an index of some description
> 
> Once I address #696 and get it merged, I'd be very happy to either write these checks, or pair with you to do so; can we try to enumerate the conditions where each is expected/required?

Yes to this - yes please!

I just feel exhausted at the idea of updating these files (i'm no good at regex) so I might need to work up the energy for it first

PR #849, opened 2024-08-03
Feature: REMOVE OLD CURRICULUM and replace with portal view

SallyMcGrath:
## What does this change?

This deletes our defunct webdev curriculum from last year.
it replaces it with basically an index page and a data folder
each course can manage their own description

>[!TIP]
> [filtered view of actual changes minus deletion of old curriculum](https://github.com/CodeYourFuture/curriculum/pull/849/files?file-filters%5B%5D=.html&file-filters%5B%5D=.scss&file-filters%5B%5D=.toml&file-filters%5B%5D=No+extension&show-viewed-files=true&show-deleted-files=false)

# Depends on

### MUST be deployed before August 17

- [x] MOST IMPORTANT: The Launch has completed AUGUST 10
- [x] Each listed module has been made into its own Hugo module  #851 
- [x] These domains have been created https://github.com/CodeYourFuture/devops_team/issues/62
- [x] All the new sites are merged and deployed
- [x] The subdomains are connected

## Checklist of sites that must exist before this is merged
https://github.com/CodeYourFuture/devops_team/issues/62
- [ ] itd.codeyourfuture.io / digital.codeyourfuture.io NONblocking
- [x] programming.codeyourfuture.io
- [ ] piscine.codeyourfuture.io #838 
- [x] sdc.codeyourfuture.io 
- [x] launch.codeyourfuture.io #851 
- [x] tracks.codeyourfuture.io #852

### Common Theme?

Just a tiny css tweak on map.scss and card.scss as it was really bugging me

Issue number: #issue-number

### Org Content?

>[!WARNING]
>Yes, this deletes ALL the old org content in this module and replaces it with a single index page

And some [data files in data/courses](https://github.com/CodeYourFuture/curriculum/pull/849/files?file-filters%5B%5D=.toml&show-viewed-files=true&show-deleted-files=false)

Each team should edit their own data file

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.MD)
- [x] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- [x] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
- [x] I have run my code to check it works
- [x] My changes follow our [Style Guide](https://curriculum.codeyourfuture.io/guides/code-style-guide)

## Who needs to know about this?
Screenshot 2024-08-03 at 12 27 48

PR #851, opened 2024-08-03
Content: The Launch module as module

SallyMcGrath:
## What does this change?

Blocking https://github.com/CodeYourFuture/curriculum/pull/849

1. created a new site, will be launch.codeyourfuture.io
2. copied module from curriculum
3. laid out like piscine, with entry, sprints, and demo
4. stubbed a cms, but cannot test until it's deployed as it's a decap cms 

### Common Content?

I added a coming soon block because there was no demo day view yet

- [ ] Block/s

### Org Content?

whole new site: org-cyf-launch
Screenshot 2024-08-03 at 13 42 08

## Checklist


kfklein15:
For some reason, my preview is showing the old SDC, not the Launch view you have the screenshot.
image
Any idea why, @SallyMcGrath ?

SallyMcGrath:
> Why do we need so many copies of the same assets and static files for each site? Are they different? Could we be using symlinks or mounting them somehow, rather than checking in a bunch of copies?

We can create a Hugo module for the static folder and mount it in if we want to. You could do this!

It's just logos really - it's really normal to have zillions of logos in a website, sorry to break it to you! 😂 

The assets are the custom overrides for each site, so they need to belong to the site: custom-images, custom-theme etc

SallyMcGrath:
> For some reason, my preview is showing the old SDC, not the Launch view you have the screenshot. image Any idea why, @SallyMcGrath ?

Because this doesn't exist as a deployed site yet. It can't until the PR is merged - as this is the first creation. We can only read the files on github or build it locally on this PR.

SallyMcGrath:
@illicitonion 

In fact I think this is how you would do it

1. move all the logos into an org-cyf assets module (NOT common theme - our logo does not belong in a white label theme)
1. mount the static folder  and assets folder directly into all the other cyf sites

It can't be any of the individual sites because the redirects are in there too

PR #861, opened 2024-08-04
Bug (tweak) move index files to docs

SallyMcGrath:
I moved the index files I had made in /common-content to common-docs/content/common-content so the pages are only created in docs and not on just any site that mounts common-content.


illicitonion (approved):
Sure

PR #868, opened 2024-08-07
fixes broken exercism python link

fcaroline2020:
## What does this change?

fixes broken exercism python link


SallyMcGrath (approved):
🫶

Issue #869, opened 2024-08-07
Clean up theme for release


SallyMcGrath:
Our theme is a bit messy and there are duplicated files and this makes @illicitonion sad.

Let's tidy up!

# What is the work that needs to be done

- [ ] move all the CYF assets to /org-cyf
- [ ] move all the mandatory assets to common-theme
- [ ] move all the required brand assets to org-cyf/assets
- [ ] move all the required static files eg logos to org-cyf/static
- [ ] [mount](https://gohugo.io/hugo-modules/configuration/#default-mounts) the org-cyf assets to each org-cyf module on a merge setting
- [ ] clean up the duplicated files in each cyf repo
- [ ] split up the common-theme config into a [Hugo config dir ](https://gohugo.io/getting-started/configuration/#configuration-directory)including a commented out guide in the per-org settings eg module.toml
- [ ] touch a [theme.toml](https://gohugo.io/contribute/themes/) in common-theme
- [ ] check everything is still hunky dory and release theme

PR #882, opened 2024-08-17
create an undeployed theme module

SallyMcGrath:
this is a module to hold all the shared CYF theme files so Daniel isn't grieved by the duplication any more

towards #869


SallyMcGrath (commented):
a temporary stage

PR #893, opened 2024-08-26
Trim leading - from objective bullet points

illicitonion:
We already have a checkbox acting as a bullet point, we don't need an additional `-` which is an artifact of being a markdown bulleted list.


SallyMcGrath:
I don't object but I'm curious - where is the dash appearing?

illicitonion:
e.g. on the Launch Apply page:

Screenshot 2024-08-27 at 12 49 14

With this PR:

Screenshot 2024-08-27 at 12 49 54

SallyMcGrath (approved):
LGTM TY

PR #921, opened 2024-09-16
Error if a module success page is missing learning objectives

illicitonion:
Also, add missing LOs so this check doesn't fail.


SallyMcGrath (approved):
LGTM

SallyMcGrath:
I'll run round to Migracode and fix theirs

PR #929, opened 2024-09-16
Define outcomes as a good job in tech

### LINKED ISSUE #927: Success criteria to be added to How our Curriculum Works
Module(s): How our Curriculum works
Week(s): N/A

The content of the page "[Success criteria"](https://docs.codeyourfuture.io/leaders/running-the-course/assessment/success-criteria) should be brought into How our Curriculum works (the bits that still make sense)

It's important so we can have all our main guides and supportive tasks for our participants and trainees in one location: the curriculum

**Additional context**

@SallyMcGrath @kfklein15  - I need to know so I can archive the CYF Docs page when done

illicitonion:
This copies over the content from
https://docs.codeyourfuture.io/leaders/running-the-course/assessment/success-criteria with some modifications - specifically:

* Trying to describe _why_ we still teach the same things to people without right to work.
* Broading "in tech" to include roles like support desk

Fixes #927

### DIFF
--- a/common-content/en/module/how-our-curriculum-works/our-mission/index.md
+++ b/common-content/en/module/how-our-curriculum-works/our-mission/index.md
@@ -6,6 +6,7 @@ facilitation = false
 emoji= '🧩'
 [objectives]
 1='Define the mission of our organisation'
+2='Define a good job in tech'
 [build]
   render = 'never'
   list = 'local'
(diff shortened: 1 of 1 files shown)


kfklein15 (approved):
Looks great. Thank you for doing this so quickly :)

PR #930, opened 2024-09-16
Move netlify.toml files

illicitonion:
They need to be in the root of the package directory, as per https://docs.netlify.com/configure-builds/monorepos/#use-a-netlify-configuration-file putting them in a tooling subdir means they get ignored.


SallyMcGrath (approved):
Ah!

PR #933, opened 2024-09-17
Don't double-navigate when pressing left/right

### LINKED ISSUE #907: Scrolling forwards through prep blocks with "right arrow" sometimes skips some
I can't reliably reproduce this, but I've seen it multiple times and on different pages.

I went to https://programming.codeyourfuture.io/how-our-curriculum-works/prep/ and pressed the right arrow on my keyboard, and was taken to "Mission" not "Background".

The same happened on https://programming.codeyourfuture.io/how-our-curriculum-works/sprints/self-educate/prep/ - I went, pressed the right arrow on my keyboard, and was taken to "No lectures" not "Flipped classroom".

Pressing the left arrow afterwards took me correctly to the n-1 page in the list, rather than back to the page I'd come from.

illicitonion:
Previously if you had focused a child element (e.g. by expanding the learning objectives), the event would get handled multiple times.

Instead, stop propagation of the event.

Fixes #907

### DIFF
--- a/common-theme/assets/scripts/solo-view.js
+++ b/common-theme/assets/scripts/solo-view.js
@@ -89,17 +89,21 @@ class SoloView extends HTMLElement {
 
   // Navigation logic
   navigateBack = (event) => {
-    event.preventDefault();
-    const backIndex = this.state.currentBlockIndex - 1;
-    this.state.tocLinks[backIndex].click();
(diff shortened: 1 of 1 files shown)


SallyMcGrath (approved):
tested vaguely on kbd and eyeballed

PR #947, opened 2024-09-18
Remove extra copies of netlify functions

illicitonion:
As of https://github.com/CodeYourFuture/curriculum/pull/946 we only deploy the function to curriculum., and all the other sites just use the curriculum. function.


SallyMcGrath (approved):
🧹

PR #964, opened 2024-09-20
942 Retro block video uses Jamboard

SallyMcGrath:
also rm sprint language just because it's used in other contexts occasionally

- [x] I have [made a FigJam](https://www.figma.com/board/otHBufZ3jGm9S7QPxeuXYV/Retro-Board-CYF?node-id=0-1&t=UEnj1KfKssCK1hYl-1) because we have these free pro accounts with Figma so we won't hit the limits on Miro and accidentally disable someone's vital board
- [x] I have added a few other people as editors eg @illicitonion @moneyinthesky @kfklein15 etc and made it public edit with a password
- [x] I have changed the video I found in wow2 to say figjam, re-rexported and [uploaded it to Youtube](https://www.youtube.com/watch?v=TeKHx6GY3Zw)
- [x] I have edited the link in the common retro block
- [x] I have searched in common to find any other references to this video and found none
- [x] I have put a note in Slack saying where the board is and how to access it

### DIFF
--- a/common-content/en/blocks/retro/index.md
+++ b/common-content/en/blocks/retro/index.md
@@ -1,7 +1,6 @@
 +++
 title="Retro: Start / Stop / Continue"
 description="Retro on the class day and the sprint"
-
 time= 25
 hide_from_overview=true
 [build]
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks!

PR #974, opened 2024-09-20
853 regex

### LINKED ISSUE #853: BUG: readme regex is over-escaping
Oh no! Looking at my preview on http://localhost:1313/common-theme/block-types/readme/ it looks like we're over-escaping...

Screenshot 2024-08-04 at 12 47 38

_Originally posted by @illicitonion in https://github.com/CodeYourFuture/curriculum/pull/846#discussion_r1703149243_

SallyMcGrath:
oh render loop!

Fixes #853 and a bunch of other little bugs 

We were losing our shortcodes and some html on the readme blocks. It's because I moved the image paths regex into a partial. Moving out of this scope and back again did something to Hugo's markdownifier, and it stopped parsing non-native markdown like shortcodes. 

I maintain this regex should be in a partial, but until we understand how to do it properly we cannot be allowed to do it at all. So I've moved the regex back into readme block.

I also in the course of this added some notes to the workshops readme at the bottom, which can also be used as a little test for this https://github.com/CodeYourFuture/CYF-Workshops

You can test images are still coming through by adding a readme or issue with an image as a block, like this for example https://github.com/CodeYourFuture/curriculum/pull/963

```toml
[[blocks]]
name="test Image"
src="https://github.com/CodeYourFuture/curriculum/issues/963"
```


README.md
```diff
@@ -36,27 +36,27 @@ To develop a specific site, `cd` into the relevant sub-directory and follow the
...
+hugo mod init github.com/{your-org}/{your-repo}/{your-module}
 ```
 
 Then add the common theme and content modules as hugo modules to hugo.toml:
 
 ```toml
 [module]
   [[module.imports]]
-    path = "github.com/CodeYourFuture/curriculum/common-theme"
+    path = 'github.com/CodeYourFuture/curriculum/common-theme'
```
illicitonion, line 54: Just curiosity: Is there a reason to prefer single quotes in this kind of context? I am a strong "double quotes by default" person (because it's more often that I want a `'` inside a quoted string than a `"`)...

illicitonion (approved):
Sure! Thanks!

README.md
SallyMcGrath: Oh no sorry that's my inattention. I changed them to test whether I needed to htmlUnescape, but I didn't, I just had vendored my modules in common-docs and wasn't getting my updates.

PR #986, opened 2024-09-22
we probably don't want more boxes in boxes

SallyMcGrath:
let's just try  knocking down the text instead for now

We can revisit this when we are not rushing to launch
Screenshot 2024-09-22 at 12 25 17


fcaroline2020 (approved):
LGTM 🚀

PR #987, opened 2024-09-22
965 Tooltips can be hidden behind header

SallyMcGrath:
position the tooltip from top instead of bottom to avoid the header
we don't really want to start trying to manipulate the z-index from inside a landmark so let's just move it down instead of up and and it will be mostly fine

Screenshot 2024-09-22 at 12 48 52


illicitonion (approved):
Hopefully we don't run into the same issue withe the footer!

PR #989, opened 2024-09-22
return accounts activity to prep

### LINKED ISSUE #812: Bring induction and setup from Induction and Fundamentals
There are many useful setup exercises in Fundamentals and Induction. They are all solving known problems, like getting people set up with accounts and software, setting expectations and culture, and explaining how to use the curriculum. 

When I drafted ITP from the old SDC, I brought over quite a few blocks but I think there are quite a few more that need bringing over.

Look in common-content/module/fundamentals or https://curriculum.codeyourfuture.io/fundamentals/)

SallyMcGrath:
we need to clearly tell them to do this or we will be back to the days of sexydino69

Fixes #812

I think the rest can go, but we do need this

### DIFF
--- a/common-content/en/module/induction/accounts/index.md
+++ b/common-content/en/module/induction/accounts/index.md
@@ -14,8 +14,8 @@ Download and install the following software or register for the apps.
 
 Click on the links to go to the registration pages. Keep your login details for all sites/apps consistent and always use your real name.
 
-{{<note type="tip" title="Real Name" >}}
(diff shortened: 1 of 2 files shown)


illicitonion:
I am not sure how to interpret this:

Screenshot 2024-09-23 at 11 44 07

Can we replace it with explanatory text, rather than trusting people will identify a pattern from this?

It's not clear to me whether this is suggesting a hierarchy of improvement, or just a list of examples.

Is JDoe better than JohnDoe? I'm not sure. If we think it is, we should explain why so people can replicate it.

Is there unsaid advice here to hide that you have a female name if you are female? If so, I think we should be explicit about it with justification, or else stop implying it.

moneyinthesky (approved):

PR #1030, opened 2024-10-03
Create a placeholder 404 file with the correct permissions

### LINKED ISSUE #992: Weird permissions problem on common-docs
**Describe the bug**

https://common.codeyourfuture.io/  / https://github.com/CodeYourFuture/curriculum/tree/main/common-docs

When I run this without vendoring the modules, I get a permissions error

```
ERROR Rebuild failed: render: failed to render pages: open /Users/me/CYF/Products/curriculum/common-docs/public/404.html: permission denied
```

I was clearly pounding through stuff and fixed this in dev by running `hugo mod vendor` but this isn't ideal. 

I don't know why this error is happening on this build only. I'd like to find out, and fix it properly. Why is this site different to the others?

**How could this be fixed?**

I don't know! But the outcome should be that we can build this site without vendoring modules.

illicitonion:
Hugo will correctly replace this with the actual rendered 404 page. But if this file doesn't already exist with the correct permissions, hugo will for some reason which we can't debug create a placeholder file with permissions 0444 instead of 0644, and then won't be able to overwrite it with the correctly rendered file.

Fixes #992 somehow, even if we don't really know how.

### DIFF
--- a/common-theme/static/404.html
+++ b/common-theme/static/404.html
@@ -1,0 +1,11 @@
+This is a placeholder file which will be replaced by hugo.
+
+If this file doesn't exist, hugo will for some reason create a placeholder file with the wrong permissions, and won't be able to overwrite it with the rendered file.
+
+You should never see this file served by a hugo site.
+
+We wish we understood why hugo was creating a placeholder with permissions 0444 instead of 0644, but we can't work it out.
+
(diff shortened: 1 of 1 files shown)


SallyMcGrath (approved):
I don't care if it works or not - this is so wonderful I approve this message 😂

PR #1033, opened 2024-10-04
Rework 24 hour clock worked example prep

### LINKED ISSUE #243: Content: Create coursework around considering input classes and edge cases
We should probably also also have a section where we talk about how this code is buggy because it assumes there are no minutes, and prompt the trainees to add a new assertion and fix that themselves. They already have all of the building-blocks they need to do this, so it should just be a matter of re-applying those same concepts around slicing.

I'd be tempted to also set a piece of coursework around considering the classes of input, and interesting edge cases, they would maybe want to test. Let's note in that exercise also that there are two major ways of identifying these classes/edge-cases: "Domain-driven" (i.e. "morning vs afternoon", "midnight and midday", etc), and "Implementation-driven" (e.g. remembering that we need to include the correct minutes isn't necessarily a core _domain_ problem, but we _can_ identify it as an interesting case to test based on knowing what our implementation looks like - we're separating the hours from minutes, so may need to pay extra attention to the idea that we may do something wrong with the minutes). This would be a great discussion for TAs to facilitate on the Saturday.

_Originally posted by @illicitonion in https://github.com/CodeYourFuture/curriculum/pull/19#discussion_r1252280431_

illicitonion:
## What does this change?

* Pull out an entire prep block about interpreting errors - we frequently see trainees ignore errors and wildly guess, this tries to guide away from that more.
* Also a block about strategies to handle errors when we run into them.
* Split "how to conditionals" and "what to do in the other branch" into two separate stages of solving the problem, and add "test and get feedback you've solved one of them" as a suggested step before solving the other one.
* Explain why we need to convert to numbers before comparing strings.
* Add a block on refactoring out repetition
* Add a block about identifying classes of input and edge cases.

I've added a prep exercise to continue the edge-case finding in https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/pull/41.

Fixes #243

### Common Content?

Yes, it reworks a lot of js1 blocks.

Hopefully anyone depending on these blocks will be broken by the fact that the `js1/assembly` block has been removed, and will see they need to update what they pull in.

### Org Content?

Sprint

## Checklist


common-content/en/module/js1/assertions/index.md
```diff
@@ -140,49 +139,21 @@ Create a javascript file on your local machine and execute the code above. Doubl
...
-}
+We need to assert that the function behaves correctly when the time is later than midday.
 
-const currentOutput = formatAs12HourClock("08:00");
-const targetOutput = "08:00 am";
-console.assert(
-  currentOutput === targetOutput,
-  `current output: ${currentOutput}, target output: ${targetOutput}`
-);
+Before we think about any code, we should think about our problem. If we think about the problem before we write the code, we can focus better - first we can focus on the data, then we can focus on the code.
```
SallyMcGrath, line 148:
```suggestion
Before we think about any code, we should think about our problem. If we think about the problem before we write the code, we can focus better - first we can focus on the data, then we can focus on the code.
```

What am I gonna say about this sentence, Daniel! 😝

common-content/en/module/js1/interpreting-errors/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+Each line of output here tells us something useful.
+
+The first line tells us exactly where the problem happened:
+
+```
+/Users/dwh/CYF/clock-example.js:12
+```
+
+It says the problem was in a file called `/Users/dwh/CYF/clock-example.js`, on line 12 (you probably had a different file name in your output). See how the filename ends with `:12` - this means "on line 12".
```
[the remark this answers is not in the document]
SallyMcGrath: So, there's a larger issue here, which is that this isn't active yet. 

I think you need to turn this into an activity where you show them how to find what line the error is on. Then have them identify two further facts from the message. At the moment you're just telling telling.

common-content/en/module/js1/reusing-variable-names/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+```
+SyntaxError: Identifier 'currentOutput' has already been declared
+```
+
+### Knowing what we changed
+
+It can be useful to remember when our code last worked, and what we changed since then.
+
+When we just had the first 10 lines of code here, everything worked, but we got this error when we added the rest:
```
SallyMcGrath, line 28:
```suggestion
When we just had the first 10 lines of code here, everything worked. When we added the rest, we got this error:
```

common-content/en/module/js1/reusing-variable-names/index.md
```diff
@@ -0,0 +1,195 @@
...
+  currentOutput === targetOutput,
+  `current output: ${currentOutput}, target output: ${targetOutput}`
+);
+```
+
+{{<note type="tip" title="Tip">}}
+
+This is one of the reasons it's useful to run our code often.
+
+If we only changed one small thing since our code last worked, it's easier to find the problem. If we wrote 100 lines of code since we last saw it work, it will be harder to identify which change was the problem.
```
SallyMcGrath, line 54:
```suggestion
If we changed one thing since our code last worked, we know what change is the problem. If we have written 100 lines of code since we last saw it work, we have 100 possible problems.
```

common-content/en/module/js1/reusing-variable-names/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+{{</note>}}
+
+### Interpreting the error message
+
+The error message is designed to tell us useful information:
```
SallyMcGrath, line 60:
```suggestion
The error message tries to tell us useful information:
```

common-content/en/module/js1/reusing-variable-names/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+```
+SyntaxError: Identifier 'currentOutput' has already been declared
+```
+
+When we get an error, we should make sure we understand all of the words in the error message. If we don't, we should look them up or ask someone.
+
+* **SyntaxError** - If we Google "JavaScript SyntaxError", [MDN tells us](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) this is "an error when trying to interpret syntactically invalid code". So - we wrote some code which isn't allowed.
```
SallyMcGrath, line 68: rephrase this whole thing as an activity where they have to do this

SallyMcGrath (changes requested):
Loving it so far.

I'm going to stop reviewing because there are two general themes which you can just run with.

1. Break up your sentences.
2. Convert 2/3 of your information into activities, retaining the first piece as a worked example and the last piece/ your summary analysis. 

Then let's go again! 🚀 ❤️ 🙌

common-content/en/module/js1/strings/index.md
```diff
@@ -53,8 +53,28 @@ time.slice(0, 2); // will access the characters below
...
+
+Think about the numbers 100 and 23. Which one is bigger?
+
+Write code which compares the strings `"100"` and `"23"` and tells you which one is bigger.
+
+Does this match your expectation? If not, what do you think is happening?
+
+{{</note>}}
+
+JavaScript uses different ways to compare values depending on their types. If you compare two strings (which may contain numbers), it will do something different than if you compare two numbers.
```
SallyMcGrath, line 70:
```suggestion
JavaScript uses different ways to [compare values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) depending on their types. If you compare two strings (which may contain numbers), it will do something different than if you compare two numbers.
```

common-content/en/module/js1/sub-goal-2/index.md
```diff
@@ -0,0 +1,60 @@
...
+[objectives]
+    1='Explain how different concepts are combined to solve a goal'
+[build]
+  render = 'never'
+  list = 'local'
+  publishResources = false
+
++++
+
+Now we can worry about what we do when we've identified a time is after midday.
```
SallyMcGrath, line 17:
```suggestion
Now we can think about what we do when we've identified a time is after midday.
```

common-content/en/module/js1/sub-goal-2/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+Earlier we observed what to do when the time goes beyond midday: **subtract 12 from the hours time to get the new hours for the 12 hour clock time**.
```
SallyMcGrath, line 19:
```suggestion
Earlier, we observed what to do when the time goes beyond midday: **subtract 12 from the hours time to get the new hours for the 12 hour clock time**.
```

SallyMcGrath (approved):
Love it! Some tiny notes

common-content/en/module/js1/interpreting-errors/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+```
+/Users/dwh/CYF/clock-example.js:12
+```
+
+Your output was probably different. But it will have the same parts. Some text, then a colon (`:`), then a number.
+
+{{<note type="exercise" title="Exercise">}}
+
+Work out what the parts of this line mean.
```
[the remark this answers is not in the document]
illicitonion: How would you recommend doing this? By manually putting heading text in each exercise? AFAICT the title attributes are unused...

common-content/en/module/js1/identifying-missing-tests/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+So we wrote another test - for the input `23:00`. We saw our implementation failed that test. And we fixed it.
+
+And we had an implementation that passed all our (2) tests!
+
+> When will we be happy that our implementation works for all possible inputs? When do we have enough tests?
+
+### Groups of input
+
+One way to approach this is to think about what _groups_ of input there may be to our problem.
```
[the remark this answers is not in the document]
illicitonion: Did a slightly different rewording.

common-content/en/module/js1/interpreting-errors/index.md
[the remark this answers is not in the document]
illicitonion: I'm going to leave this for now so I can merge, as this sprint is next sprint. Happy to come back and add these after.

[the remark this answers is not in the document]
SallyMcGrath: Just 1. 2. 3.  inside the text

PR #1037, opened 2024-10-06
add calendar link to HTW

SallyMcGrath:
## What does this change?

Added the calendar to how this works

### Common Content?

Yeah I added a block. I made it separate as other orgs don't want our calendar. When we move all this into a module, probably this block would be locally overriden.


SallyMcGrath:
just gonna force this through as things are kind of 🔥

PR #1040, opened 2024-10-08
Unify on just one How This Works module (from ITP)

### LINKED ISSUE #922: How Our Curriculum Works is inconsistently copied across multiple sites
`{itp,sdc,tracks}/content/how-our-curriculum-works/success/index.md` currently contain no learning objectives, so I'm adding them. But I don't want to have to add (and maintain) the same LOs across each version. I was about to symlink across between them, but it feels like ideally there would be one common version that all three sites reference, rather than picking one to be canonical and having the others reference it.

Similarly, `org-cyf-itp/content/how-our-curriculum-works/sprints/self-educate/prep/index.md` is currently missing the `module/how-our-curriculum-works/dialogue` block between No Lectures and Importance of Prep - the other two sites have this. Ideally these would be structurally shared as the same, rather than copied and kept in sync.

@SallyMcGrath What's the best way to achieve this?

illicitonion:
## What does this change?

Currently we have several copies that are intended to be in sync.

If we want to have them diverge in the future, we can do that.

Also, run lots of `hugo mod tidy`, and clean up the resulting go.mod files.

Fixes #922

### Org Content?

Module

## Checklist


SallyMcGrath (approved):
Great - thank you - let me know the pain points when you get chance

illicitonion:
@SallyMcGrath Honestly not much by way of pain-points. I think the biggest thing was needing to know to run `hugo mod init`, seeing it fail (because the remote resource didn't exist yet), then needing to know to add the override.

PR #1059, opened 2024-10-10
Add brief intro to anonymous and arrow functions

illicitonion:
## What does this change?

We started to use anonymous functions in Jest, and we're going to start using arrow functions in the next sprint. These aren't scary things, so we simply explain that they exist and are the same as each other.

We even do so _after_ using them in Jest, rather than before, to emphasise that they're not scary, and that you probably understood what they were doing when reading the examples without realising.

### Common Content?

Yes - adds two new blocks

- [ ] module/js1/anonymous-functions/index.md
- [ ] module/js1/arrow-functions/index.md

### Org Content?

Sprint - adds two blocks to org-cyf-itp/content/structuring-data/sprints/3/prep

## Checklist


common-content/en/module/js1/arrow-functions/index.md
```diff
@@ -0,0 +1,60 @@
...
+[build]
+  render = 'never'
+  list = 'local'
+  publishResources = false
+
++++
+
+As we write more code, we are going to write lots and lots of anonymous functions.
+
+JavaScript added even shorter ways of writing an anonymous function. These four functions all do the same thing:
```
SallyMcGrath, line 18:
```suggestion
JavaScript has even shorter ways of writing an anonymous function. These four functions all do the same thing:
```

common-content/en/module/js1/arrow-functions/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+```js
+// We can skip the name of the function if we don't need it to have a name.
+function (decimalNumber) {
+  return `${decimalNumber * 100}%`;
+}
+```
+
+```js
+// We can also skip the keyword function.
```
SallyMcGrath, line 34:
```suggestion
// We can also skip the keyword 'function'.
```

common-content/en/module/js1/arrow-functions/index.md
SallyMcGrath: I think you might be inadvertently telling them that arrow functions are anonymous functions here. Can you think of how to mitigate this risk?

common-content/en/module/js1/arrow-functions/index.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+```js
+// If our function just returns a single value,
+// without needing any other statements in our function,
+// we can even skip the return keyword.
+(decimalNumber) => `${decimalNumber * 100}%`;
+```
+
+This can be useful to make is quicker and easier to write functions. It also reduces the number of things we need to skip past when reading a function.
```
SallyMcGrath, line 48:
```suggestion
This can make it easier and quicker to write functions. It reduces the number of things we need to read in a function. 
```

SallyMcGrath (changes requested):
A few notes as I see them -- nothing intensive

common-content/en/module/js1/arrow-functions/index.md
illicitonion: Technically they are, aren't they? An arrow function _is_ anonymous until it's bound to an identifier?

```js
// This function is anonymous, it just happens to be defined as an arrow function.
test("foo" () => {});

// This only stops being anonymous because it happens to be bound to an identifier.
// The fact that it's an arrow function is incidental.
const testFunction = () => {};
test("bar", testFunction);
```

common-content/en/module/js1/arrow-functions/index.md
illicitonion: I'm keeping the also unless you're going to fight me over it - reading and writing are different concerns, and making clear the second sentence is adding to, not clarifying, the first feels useful.

common-content/en/module/js1/anonymous-functions/index.md
```diff
@@ -0,0 +1,57 @@
...
+We can imagine the `test` function is defined like this:
+
+```js
+function test(name, testFunction) {
+  // Call the passed test function
+  testFunction();
+}
+```
+
+It doesn't matter what _we_ named our test function - inside the `test` function it has the name `testFunction`, so Jest can call it.
```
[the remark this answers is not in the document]
illicitonion: Slightly reworded, but kept the gist (2c99901047e97416a40f3d7b78955ca2b7a3bb45).

common-content/en/module/js1/arrow-functions/index.md
SallyMcGrath: It is, but what I think could happen is that trainees who don't see a few examples of  `const namedFunction = () => ` could model this as arrow functions _are_ anonymous, as in this is a feature of arrow functions not of all functions.

illicitonion: Added a clarifying blurb in aac5e1a1718f371bfb159c41526841eb77f8037f

common-content/en/module/js1/arrow-functions/index.md
```diff
@@ -0,0 +1,60 @@
...
+(decimalNumber) => {
+  return `${decimalNumber * 100}%`;
+}
+```
+
+```js
+// If our function just returns a single value,
+// without needing any other statements in our function,
+// we can even skip the return keyword.
+(decimalNumber) => `${decimalNumber * 100}%`;
```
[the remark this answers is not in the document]
illicitonion: If you find it, come add it later! :)

common-content/en/module/js1/arrow-functions/index.md
```diff
@@ -0,0 +1,60 @@
...
+[objectives]
+    1='Write an arrow function'
+[build]
+  render = 'never'
+  list = 'local'
+  publishResources = false
+
++++
+
+As we write more code, we are going to write lots and lots of anonymous functions.
```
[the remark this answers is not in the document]
illicitonion: Added in aac5e1a1718f371bfb159c41526841eb77f8037f - it's a little technical, but given we literally just covered it in the previous block, I think that's actually an ok reinforcement.

common-content/en/module/js1/anonymous-functions/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+Because it doesn't matter what we named the function (because we never call it by name), we didn't give it a name.
+
+Otherwise, these two functions act the same. The only difference between them is whether we created a variable name for the function in the scope where we defined it.
```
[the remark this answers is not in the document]
illicitonion: Added in aac5e1a1718f371bfb159c41526841eb77f8037f

common-content/en/module/js1/anonymous-functions/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+{{<note type="exercise" title="👀 Spot the difference">}}
+
+Stop and identify the difference in syntax between these two function definitions.
+
+{{</note>}}
+
+We didn't give a name to the function in our Jest test.
+
+This is ok, because we don't need it to have a name. We don't call the function by name. We passed the function as an argument to the `test` function. The `test` function takes the function as a parameter. And function parameters get their own names in the scope of the function.
```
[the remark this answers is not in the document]
illicitonion: Added in aac5e1a1718f371bfb159c41526841eb77f8037f

common-content/en/module/js1/anonymous-functions/index.md
```diff
@@ -39,7 +39,7 @@ Stop and identify the difference in syntax between these two function definition
 
 We didn't give a name to the function in our Jest test.
 
-This is ok, because we don't need it to have a name. We don't call the function by name. We passed the function as an argument to the `test` function. The `test` function takes the function as a parameter. And function parameters get their own names in the scope of the function.
+This is ok, because we don't need it to have a name. We don't call the function by name. We passed the function as an {{<tooltip title="argument">}}Arguments are values given to a function which can be different every time we call the function.{{</tooltip>}} to the `test` function. The `test` function takes the function as a {{<tooltip title="parameter">}}A parameter is a named variable inside a function. The variable's value is given by the caller, when the function is called.{{</tooltip>}}. And function parameters get their own names in the {{<tooltip title="scope">}}Scope is where a variable can be accessed from. When we define function, its parameters are only available inside the function.{{</tooltip>}} of the function.
```
SallyMcGrath, line 42: This is a thought that I don't know the answer to so I invite you to think it too.

We have defined all these terms before, eg 

https://programming.codeyourfuture.io/structuring-data/sprints/1/prep/#parameters 
A parameter is a special kind of variable: its value is defined by the caller.
🧶 argument
Arguments are inputs given to a function inside (). An argument means an input.

https://programming.codeyourfuture.io/structuring-data/sprints/1/prep/#scope

etc

Do we think it's useful to always use the same def, or keep complicating it, or some mixture of the two?

SallyMcGrath (approved):
LGTM - something to think about for the future in notes

common-content/en/module/js1/anonymous-functions/index.md
illicitonion: Yeah, I've have this thought too and am pretty conflict about it.

I think in general being able to hark back to a "general purpose definition" is useful in 80% of the cases (and when this happens, I exactly copy+paste tooltips and am sad it's unmaintained), but I consciously rewrote these rather than copying them because I thought it was more clear in context (and that there's hopefully more understanding now than when we first introduced them)...

Basically I think there are three contexts for a definitional tool-tip:
1. General purpose
2. Simplified because it's a first introduction and needs simplification.
3. Special-cased because something specific is being pointed out / referenced that wouldn't _generally_ be useful in a definition.

Which leans me to: It would probably be useful to have a central glossary-ish thing somewhere that can be invoked through a shortcode, and I'd probably use it 80% of the time, but it doesn't feel as high priority a thing to build as it would if I would use it 99% of the time...

illicitonion: Forked off discussion into https://github.com/CodeYourFuture/curriculum/issues/1083

PR #1111, opened 2024-10-22
Non-trivial tweaks to Data Flows sprints 2 and 3

illicitonion:
## What does this change?

As well as a bunch of copy edits, a few non-trivial changes:
* Sprint 2's code didn't actually fully work in a bunch of ways. Add soem blocks and steps making sure we actually step through all the issues that come across in the code we're teaching. I'm consciously leaving this mostly as a worked example, rather than do-yourself exercises, because the trainees already have the TV Show project as an entire do-yourself exercise about the same topics.
* When working with promises, consistently tell people to save examples to files and run them, rather than run them in the repl. This is because if you execute these snippets in a repl you get printed out extra Promise values (because the repl prints out each evaluated expression), which makes it much harder to do things like "spot the difference between what returns a promise and what evaluates to a value").
* Completely remove `Promise.catch` This is probably the most controversial but I think it's important. Two main reasons:
   1. We haven't actually taught `try-catch` anywhere, so we can't just rely on transferring "it's the same as you already know, just async".
   2. There are already _way_ too many novel concepts in Data Flows Sprint 3, and it's already very hard to follow from scratch. If we add `try-catch` earlier in the course (which I think we should), we should re-add this.

### Common Content?

Yes. Much of js3. Including adding new blocks.

### Org Content?

Yes - Data Flows sprints 2 and 3 prep, and sprint success.

## Checklist

### DIFF
--- a/common-content/en/module/js3/identifying-state/index.md
+++ b/common-content/en/module/js3/identifying-state/index.md
@@ -1,7 +1,7 @@
 +++
 title = '🔎 Identifying state'
 
-time = 15
+time = 20
 facilitation = false
 emoji= '🧩'
 [objectives]
@@ -33,11 +33,13 @@ In our film example, we would not store "is the search term empty" and "what is 
 
 #### 🖇️ If two things always change together, they should be one piece of state.
 
-If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}
+If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}. We would also not have a state for "background colour" - we would derive that from the one boolean.
+
+#### State in our example
 
 In our film example, we need two pieces of state:
 
 1. Our list of all films
 2. The search term
 
-When we introduce filtering films based on the search term **we will not introduce new state**. Our filtered list of films can be _derived_ from our existing state.
+When we introduce filtering films based on the search term **we will not introduce other new state** like a filtered list of films. Our filtered list of films can be _derived_ from our existing state.
--- a/common-content/en/module/js3/identifying-state/index.md
+++ b/common-content/en/module/js3/identifying-state/index.md
@@ -33,7 +33,7 @@ In our film example, we would not store "is the search term empty" and "what is 
 
 #### 🖇️ If two things always change together, they should be one piece of state.
 
-If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}. We would also not have a state for "background colour" - we would derive that from the one boolean.
+If a website allows log-in, we would not have one state for "is a user logged in" and one state for "what user is logged in". We would have one piece of state: The currently logged in user. We would set that state to `null` if there is no logged in user. We can answer the question "is a user logged in" by checking if the currently logged in user is `null`. We don't need a separate piece of state.
 
 #### State in our example
 
@@ -42,4 +42,4 @@ In our film example, we need two pieces of state:
 1. Our list of all films
 2. The search term
 
-When we introduce filtering films based on the search term **we will not introduce other new state** like a filtered list of films. Our filtered list of films can be _derived_ from our existing state.
+When we introduce filtering films based on the search term **we will not introduce other new state**. We will not store a filtered list of films in state. Our filtered list of films can be _derived_ from our existing state.
--- a/common-content/en/module/js3/introducing-new-state/index.md
+++ b/common-content/en/module/js3/introducing-new-state/index.md
@@ -4,8 +4,6 @@ title = '🆕 Introducing new state'
 time = 30
 facilitation = false
 emoji= '🧩'
-[objectives]
-    1=''
 [build]
   render = 'never'
   list = 'local'
@@ -41,7 +39,7 @@ const state = {
 
 We needed to pick an initial value for this state. We picked the empty string, because when someone first loads the page, they haven't searched for anything. When someone types in the search box, we will change the value of this state, and re-render the page.
 
-We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will be empty, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box.
+We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will initially be empty, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box in the UI.
 
 This is because we have split up our problem into three parts:
 
--- a/common-content/en/module/js3/introducing-new-state/index.md
+++ b/common-content/en/module/js3/introducing-new-state/index.md
@@ -39,7 +39,7 @@ const state = {
 
 We needed to pick an initial value for this state. We picked the empty string, because when someone first loads the page, they haven't searched for anything. When someone types in the search box, we will change the value of this state, and re-render the page.
 
-We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will initially be empty, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box in the UI.
+We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will be empty at first, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box in the UI.
 
 This is because we have split up our problem into three parts:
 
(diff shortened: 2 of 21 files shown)


common-content/en/module/js3/re-rendering/index.md
```diff
@@ -0,0 +1,95 @@
...
+        ],
+        searchTerm: "",
+      };
+
+      const template = document.getElementById("film-card");
+      const createFilmCard = (film) => {
+        const card = template.content.cloneNode(true);
+        // Now we are querying our cloned fragment, not the entire page.
+        card.querySelector("h3").innerText = film.title;
+        card.querySelector("[data-director]").innerText = `Director: ${film.director}`;
```
SallyMcGrath, line 63:
```suggestion
        card.querySelector("[data-director]").textContent = `Director: ${film.director}`;
```

common-content/en/module/js3/re-rendering/index.md
```diff
@@ -0,0 +1,95 @@
...
+        card.querySelector("time").innerText = `${film.duration} minutes`;
```
SallyMcGrath, line 64:
```suggestion
        card.querySelector("time").textContent = `${film.duration} minutes`;
```

common-content/en/module/js3/re-rendering/index.md
```diff
@@ -0,0 +1,95 @@
...
+        card.querySelector("[data-certificate]").innerText = `Certificate: ${film.certificate}`;
```
SallyMcGrath, line 65:
```suggestion
        card.querySelector("[data-certificate]").textContent = `Certificate: ${film.certificate}`;
```

Just because that's what it is on the video tbh

common-content/en/module/js3/then/index.md
```diff
@@ -48,12 +46,12 @@ It's a similar idea as the event loop we have already investigated, but this tim
...
 {{</note>}}
 
-We can chain multiple `.then()` calls to run more logic, passing the resolved value to the next callback in the chain. This allows us to handle the asynchronous response in distinct steps. Let's create a getProfile function which we can try out in our Node REPL:
+We can chain multiple `.then()` calls to run more logic, passing the resolved value to the next callback in the chain. This allows us to handle the asynchronous response in distinct steps. Let's create a getProfile function in a file, call it, and try running the file with node:
 
 ```js
 const getProfile = (url) => {
   return fetch(url)
-    .then((response) => response.json()) // This callback consumes the response and parses it as JSON into an object.
+    .then((response) => response.json()) // This callback consumes the response string and parses it as JSON into an object.
```
SallyMcGrath, line 54: VERY nice addition

common-content/en/module/js3/introducing-new-state/index.md
```diff
@@ -41,7 +39,7 @@ const state = {
 
 We needed to pick an initial value for this state. We picked the empty string, because when someone first loads the page, they haven't searched for anything. When someone types in the search box, we will change the value of this state, and re-render the page.
 
-We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will be empty, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box.
+We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will initially be empty, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box in the UI.
```
SallyMcGrath, line 42:
```suggestion
We could pick _any_ initial value. This actually allows us to finish implementing our render function _before we even introduce a search box into the page_. In real life, our `searchTerm` state will be empty at first, but we can use different values to help us with development. We can make the page look like someone searched for "Pirate", even before we introduce a search box in the UI.
```

common-content/en/module/js3/identifying-state/index.md
```diff
@@ -33,11 +33,13 @@ In our film example, we would not store "is the search term empty" and "what is
...
+
+#### State in our example
 
 In our film example, we need two pieces of state:
 
 1. Our list of all films
 2. The search term
 
-When we introduce filtering films based on the search term **we will not introduce new state**. Our filtered list of films can be _derived_ from our existing state.
+When we introduce filtering films based on the search term **we will not introduce other new state** like a filtered list of films. Our filtered list of films can be _derived_ from our existing state.
```
SallyMcGrath, line 45:
```suggestion
When we introduce filtering films based on the search term **we will not introduce other new state**.  We will not store a filtered list of films in state. Our filtered list of films can be _derived_ from our existing state.
```

common-content/en/module/js3/identifying-state/index.md
```diff
@@ -33,11 +33,13 @@ In our film example, we would not store "is the search term empty" and "what is
 
 #### 🖇️ If two things always change together, they should be one piece of state.
 
-If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}
+If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}. We would also not have a state for "background colour" - we would derive that from the one boolean.
```
SallyMcGrath, line 36:
```suggestion
If our website had light mode and dark mode, we would not have one state for "is dark mode enabled" and one state for "is light mode enabled". We would have one piece of state: a {{<tooltip title="boolean">}}true or false. On or off.{{</tooltip>}}. 

We would also not have a state for "background colour". We would derive that from the single boolean: light: TRUE/FALSE.
```

I'm pondering this. I mean it's true - that's how you write dark modes HOWEVER in a browser you always have two competing states -- there's the browser state: prefers dark mode, and then there's the override button, which you want users to have because there's always a significant minority who, for example, prefer darkmode generally but not YOUR darkmode, or vv. And there isn't a default position to start.

Therefore if they Inspect the curriculum, or any reasonably complex site, they will likely find two state settings handling dark/light, both `@media(prefers-color-scheme: dark) ` or light, to set up the initial colours, and then a class is-light-mode / is-dark-mode to add specificity to override the default preference on user selection.

And of course if they Inspect poorly written sites, of which there are many, they will see things like `.is-dark-mode {background: var(--dark-background-color);}` which is obviously heinous and mad but common. 

So is this example clarifying or potentially confusing? (I'm aware you're elaborating on an existing example --and it's the existing example I am questioning here, piqued by your elaboration)

SallyMcGrath (commented):
I'm just dropping a few comments here from earlier. We have discussed this on Slack, and I want to go through this in more detail. 

I don't have strong objections to this and think basically yes go ahead -- don't let me block you. And thank you for this!! 

I just feel there's something pinging me about this -- connected to Mitch's walk throughs which I know trainees value highly. I think probably videos are the place for exposing and resolving roadblocks, not more text. 

I think we are both too inclined to text and could do with a perspective that likes video walkthroughs.

common-content/en/module/js3/identifying-state/index.md
illicitonion: I think this is getting pretty in the weeds. But I take the point. I replaced this example with "is a user logged in" + "who is currently logged in" - does that work for you?

illicitonion:
### General "read vs do" approach

I definitely struggle with the balance between provoking learners to do things for themself _and also_ explain things to make sure they're not lost _and also_ be able to foresee issues they may run into. Would love to discuss general strategies here!

### Videos

> I think probably videos are the place for exposing and resolving roadblocks, not more text.

I'm not sure I agree _videos_ are the place for this. I think _interactive sessions_ are.

Videos can be great alternate ways of _presenting_ information, but I don't think are typically great ways of _clarifying misunderstandings_.

I can't speak for anyone else, but for me I don't typically learn much from watching someone not understanding something and then have a conversation helping them understand it. I learn from raising my own misunderstandings and having them actively resolved, or from watching other people in a dialogue and being able to chip in my own issues.

Watching someone else's recoded session of this, I:
1. Struggle to focus
2. Get bored by people debugging an issue I do understand so stop paying attention
3. Maybe don't get my own misunderstandings addressed at all
4. If they are addressed, maybe the explanations don't land with me

I think providing videos of this process _is_ better than not doing so. But for me the real value we should be offering is sessions with people who do understand (either on Saturdays, or in the week) to interact with live.

I don't think the goal of Mitch's videos was to replace these discussions, though they do serve as an "in case of emergency" backstop in case there are no volunteers around. In my idea world, learners would be able to read the material, do the exercises, and get help when they need. I think this was what Mitch was offering in the sessions when they were being filmed, and we should be wary how extensively we re-purpose them.

SallyMcGrath:
I agree they shouldn't replace discussions, but they should replace long divergence into common problems the learner will encounter. With timestamped chapters, this can really help resolve blockers outside of class day. 

I think we both need to valorise videos a bit BECAUSE we both hate them. Trainees love them (not all, but many). It's the most consistent feedback -- more videos, more videos, they love the videos. 

I can't pay attention to a video for more than a minute or so. I can't remember anything anyone says to me either! I never learn anything in a workshop or training session. I like to read some long clear narratives, sketch out the model that's built in my head, build some code to apply it, and fill in the gaps with docs. That's my preference, _massively_.  But it's not at all common, and if it's just you and me we will trend text heavy (with some pictures from me). 

Length and anticipatory explanation are your main risks. Abstraction and analogy are mine.

I think we need to keep the video-loving needs of other learners in view.

SallyMcGrath (approved):
Let's pick up our videos, explanations, blockers, activities discussion more generally and not block this.

But remember Migracode!

PR #1117, opened 2024-10-23
return assembly

SallyMcGrath:
other people are using it -- let's not silently remove content when we've no need to -- this would silently break MCB

if we're going to have common content we need to preserve it -- if there are really significant rewrites that make old versions unusable it's probably time to make a new block not rewrite


SallyMcGrath:
mergin because MCB will deploy in 30 mins

Issue #1119, opened 2024-10-24
Feature: The next class is link


SallyMcGrath:
**Which module(s) and week(s) does this change affect?**
ITP all

**What is the work that needs to be done?**

Replace "Here’s how you find out what to do when" with a partial that computes the next likely day plan and provides a link directly to it

Here's how, I guess

Given we have course data in org-cyf-theme
And this shows duration, frequency, starts, eg 
```toml
name="Intro to Programming"
code="itp"
weight=2
menu="start here"
url="https://programming.codeyourfuture.io"
description="Programming with JavaScript, Python, and SQL; collaborate to deliver working software with tests"
emoji="🐣"
days=84
commitment="part time"
frequency=4
starts=[1,4,7,10]
```
And we have the time now `time.Now`

And we add our scheduled modules to a menu called something like, er `scheduled`

Then we should be able to do something like this:

1. Find the day today `$today`
1. Find next Saturday `$nextClass`
1. Range over `starts` and find the last course that is less than or eq the month of next Saturday to set `$currentCourse`
1. Given that each course starts on the first of the month, derive the sprint number we are on now based on the distance between now and the first of the month of `$currentCourse`
1. Range over the items in the menu scheduled  and get a slice of all the dayplans in order - the page objects
1. Match the dayplan index to the sprint number to access the `RelPermalink` on the page object

Handle edge cases eg Christmas, but don't spend too long on it BECAUSE this should come from @CodeYourFuture/class-planner-devs API in the future when we have consumable APIs as they already are doing this manual work. 

BTW this is just off the top of my head so if you see a better/faster/sweeter way to do this please override me! 

**Why is this work important to do?**

Volunteers should prepare, but when they don't let's not disrupt class orienting them -- let's put a link on the front page

**Additional context**

**Who might need to know about this change?**

Mohammed-Mutee:
i am interested to contribute can guide me doing this ?

PR #1127, opened 2024-10-24
Suggested edits to discuss

illicitonion:


SallyMcGrath:
You've made it longer you monster! 😂

illicitonion:
Yeah, I agree with the one-screen goal. PTAL!

SallyMcGrath:
> Yeah, I agree with the one-screen goal. PTAL!

Hmmm, no. Actually I think we need to explicitly say start here, then do this. I think this because the curriculum already shows this without this explanation and some people don't get it. So let's try to say it. 

I think I'm gonna take your word choices and my formatting choices and call that a good first go. Link where my fabled next class partial will go for now? ( #1119 )

Issue #1105, opened 2024-10-21
Include more theme/heading information on index pages


illicitonion:
On https://cyf-itd.netlify.app/steps/ - just listing "one", "two", etc isn't very informative:

<img width="829" alt="Screenshot 2024-10-21 at 13 46 52" src="https://github.com/user-attachments/assets/e0a69286-f137-4503-b93e-4c6f9c19e6ae">

If these read "One - Make a Cover Letter with Generative AI", "Two - Evaluate data with Google Sheets" and similar, this would be a much more useful and navigable page. This information already exists in the "description" frontmatter for the steps.

The same is true of our sprint listings, e.g. on https://programming.codeyourfuture.io/user-data/

<img width="1342" alt="Screenshot 2024-10-21 at 13 48 10" src="https://github.com/user-attachments/assets/a5dc5939-e0f8-4156-8991-eadcd66796ac">

Showing "Sprint 1 - Version Control" would be more useful to understand what each sprint is. For sprints, we already have a "theme" piece of frontmatter that is designed to be a concise summary like this.

illicitonion:
Yes, I'm not saying the information doesn't exist, but that on these particular pages it may be useful to also show it?

Though also, it may be useful to show this information _without_ requiring a hover on https://cyf-itd.netlify.app/ - for the other courses we already use more descriptive names for the top-level modules (and have quite verbose descriptions which would be clutter to show by default), whereas for ITD "three" is very not-descriptive, and all of the descriptions are actually quite concise...

SallyMcGrath:
I think a better solution would be to suppress this page and skip the navigation directly back to home.

illicitonion:
Sounds like a plan! How do we do that? :)

To recap the action items I think are useful:
* ITD: Suppress https://cyf-itd.netlify.app/steps/ and make the nav go back to home
* ITD: Rename the steps to include their description in their title
* Everything except ITD: Include the theme on sprint views like https://programming.codeyourfuture.io/user-data/

PR #1131, opened 2024-10-24
1105 Include more theme/heading information on index pages

SallyMcGrath:
Addresses #1105 : ITD: Rename the steps to include their description in their title

They already have this information, so I just showed it on the page


illicitonion (approved):
Nice!

PR #1130, opened 2024-10-24
1105 Include more theme/heading information on index pages

SallyMcGrath:
PR 1 of 3

Fixes  https://github.com/CodeYourFuture/curriculum/issues/1105#issuecomment-2428695177

brings theme onto module index view

PR #1135, opened 2024-10-24
1105 suppress step view

SallyMcGrath:
Part 3 of #1105 

Other parts #1130  #1131 

1. suppress step index view as it's just the same info as the home page
2. skip this inaccessible page in breadcrumbs

### DIFF
--- a/common-theme/layouts/partials/breadcrumbs.html
+++ b/common-theme/layouts/partials/breadcrumbs.html
@@ -26,16 +26,18 @@
           </a>
         </li>
       {{- end }}
-      <li class="c-breadcrumbs__item">
-        <a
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Perfect - thanks!

Is there a reason we're using the string "never" rather than a boolean here? Other conditional-ish states you're imagining adding?

SallyMcGrath:
> Perfect - thanks!
> 
> Is there a reason we're using the string "never" rather than a boolean here? Other conditional-ish states you're imagining adding?

Render always has a value and it's 1 of 3 options:  https://gohugo.io/content-management/build-options

I feel like the only state we're sure about is never show a link to a page that never exists, so I specified it.

PR #1139, opened 2024-10-27
1001 Bug: readme links not being cleaned

### LINKED ISSUE #1001: Bug: readme links not being cleaned
**Describe the bug**

Linking to readmes used to be forgiving - you could tree, or blob, or add readme to the end and it worked out what you meant. But now it only works on one specific path.

This is a problem because GitHub provides a bunch of different urls and it's not clear which should be the canonical one so probably we should accept all. 

**How could this be fixed?**

Probably a new strings util needs to be made in https://github.com/CodeYourFuture/curriculum/tree/main/common-theme/layouts/partials/strings

And the current regexes replaced around here 

https://github.com/CodeYourFuture/curriculum/blob/874907ac6f636e00e745ae1b79c6c429957e0eef/common-theme/layouts/partials/block/data.html#L62

And then these versions of readme links should be acceptable 

https://github.com/CodeYourFuture/CYF-Workshops/blob/main/readme.md
https://github.com/CodeYourFuture/CYF-Workshops/readme.md
also tree ALSO ALSO

not pin to the branch main? There's no reason to suppose readmes are always on a branch called main at all. 

Extra points for warning when there's nothing at the endpoint but don't fail the build because we don't want non-devs to fail our build all the time because sometimes I need to sleep and also eat food.

SallyMcGrath:
## What does this change?

As per #1001, adds a new strings util to transform github readme api paths. 

I've changed the format of a link on ITP> UFD > 1 > Dayplan  to show this. This is how it worked before we broke it at some point, but I've done it in a more robust way now - I've hived it off into a separate, well commented partial that the data partial can call. 

It is probably desirable to do this with most of the string transforms over time.

### Common Theme?

 Fixes #1001 readme regex

Issue number: #issue-number

### Org Content?

UFD | 1 | Dayplan | Workshop

## Checklist

### DIFF
--- a/common-theme/layouts/partials/block/data.html
+++ b/common-theme/layouts/partials/block/data.html
@@ -56,11 +56,7 @@
       {{ $newSrc = print $newSrc "?per_page=5&state=open" }}
       {{ .Scratch.SetInMap "blockData" "type" "pullreq" }}
     {{ else }}
-      {{/* For repositories and individual README files, use readme API endpoint
-        https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-a-repository-readme-for-a-directory
-      */}}
-      {{ $newSrc = replace $newSrc "/tree/main" "/readme" }}
-      {{ $newSrc = replace $newSrc "/blob/main/README.md" "/readme" }}
+      {{ $newSrc = partial "strings/readme-paths" (dict "src" $newSrc) }}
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Seems reasonable - thanks!

SallyMcGrath:
> Seems reasonable - thanks!

I hate regex so much, Daniel! I can't understand it no matter how hard I try. So I need to write it out in excruciating detail every time. 🫠

PR #1144, opened 2024-10-28
1058 Page nav within a module doesn't see sprints

### LINKED ISSUE #1058: Page nav within a module doesn't see sprints
On https://programming.codeyourfuture.io/data-groups/prep/ the next step in the nav after "prep" is listed as "success"

In reality, it should be Sprint 1.

Screenshot 2024-10-10 at 13 28 22

SallyMcGrath:
this didn't work when I tried it before, but there's obviously something wrong in my pages model because it works now, and this is a much much smaller change

I think I prefer this to #1143 as it's just much less upheaval. Thoughts @illicitonion?

Fixes #1058

### DIFF
--- a/common-theme/layouts/_default/sprint.html
+++ b/common-theme/layouts/_default/sprint.html
@@ -14,26 +14,4 @@
       {{ end }}
     </ol>
   </article>
-  <footer class="c-page-footer">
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
I'll take it!

PR #1146, opened 2024-10-29
Update links to ITD

illicitonion:
## What does this change?

Update links to ITD

It's hosted on a proper subdomain now.

### Org Content?

Just some links


SallyMcGrath (approved):
Whoop! TY

PR #1150, opened 2024-10-31
final fixes from call with Karen

SallyMcGrath:
## What does this change?

- [ ] update all extend times to 5]reading time not doing time
- [ ] also fix english test time to stated length
- [ ] update sharelink instructions to wording agreed
- [ ] fix erroneous 5 links claim


illicitonion (approved):
Works for me! I am not sure long-term about expressing open-ended stretch tasks as "5 minutes", but that's not a question for today.

PR #1167, opened 2024-11-01
Error if a README couldn't be found

illicitonion:
GetRemote returns nil not an error in this case

Also, fail the build on unrecognised block types.


SallyMcGrath (approved):
LGTM

PR #1170, opened 2024-11-04
Introduce metadata about track-types

illicitonion:
## What does this change?

As per our last syllabus team meeting, we think these are the three families of tracks we have.

Explain a bit of context as to what each category is. This maybe makes more sense to do under the headings of each group of tracks? But then the text is going to be awfully skinny...

Also, self-study tracks may have specific prereqs - we should list them so people know what they may want to opt to study.

### Common Theme?

- [x] Adds the ability to show prereqs on module cards
- [x] Needs some styling on the card display

### Org Content?

Updates tracks metadata.


SallyMcGrath:
Oh Daniel do you want to break my heart! Absolutely not! 😂 😭 

I mean, I completely agree with the organisation and the descriptions, but we can't have this layout I will DIE. I think it might another case for Taxonomies?

And we can put some kind of info dump underneath if we need to.

illicitonion:
I've pulled out the "adding a taxonomy map" into https://github.com/CodeYourFuture/curriculum/pull/1187 and will pick up the "adding metadata to the terms" piece here when that is merged.

SallyMcGrath:
Humbly apologise for blocking this!

SallyMcGrath:
> I've pulled out the "adding a taxonomy map" into #1187 and will pick up the "adding metadata to the terms" piece here when that is merged.

Hi

I've done the layout changes and updated this branch from main. The question to answer is what specific condition do we want to show the metadata - I think it's when there's a term page, but reasonable people may differ.

common-theme/layouts/partials/map-common.html
```diff
@@ -1,34 +1,52 @@
...
-              {{ partial "card.html" $module }}
-          </li>
-        {{ end }}
+  <h1 class="e-heading c-map__start is-none--lt-container">👉🏾</h1>
+  {{ range $i, $section := index . "sections" }}
+    {{ $sectionTitle := index $section "title" }}
+    {{ $sectionTermPageLink := index $section "termPageLink" }}
+    {{ $sectionValues := index $section "values" }}
+    <section class="c-map__block">
+      {{ if $sectionTermPageLink }}
```
SallyMcGrath, line 15: I think if it's a term we flip to this disclosure view and also link to the term

SallyMcGrath: But as per comment I'm not wedded to this!

SallyMcGrath (approved):
See comments and notes!

PR #1173, opened 2024-11-06
Suggest retros in RetroTool not FigJam

### LINKED ISSUE #1104: Work out permissions for retro figjam
Every single day-plan ends with a retro block which says folks should use FigJam: https://programming.codeyourfuture.io/user-data/sprints/1/day-plan/#retro

I tried doing this in class this week, and we found:
1. Every participant needed to sign up for an account. This is a one-time cost, but was a bit annoying in class.
2. Participants were unclear as to whether they were meant to sign up for an education account, as this form was shown to them. I'm guessing we probably shouldn't be doing this until they have student emails?
3. Every participant needed to request access to the FigJam I'd created (which doesn't even get flagged in the page - I got emails about each request).

We ended up giving up and doing our retro on Slack.

If we're going to use FigJam, we should work out:
1. What up-front steps do we want participants to do (e.g. by having them set up a Figma account in https://programming.codeyourfuture.io/user-data/prep/#professional-accounts)
2. What per-retro steps are needed (e.g. assigning permissions), and if there's a way to make that smoother (e.g. maybe there's a group we can add folks to or something so that they automatically get permission).

cc @SallyMcGrath who set up the FigJam stuff originally

illicitonion:
## What does this change?

Suggest retros in RetroTool not FigJam

Figma requires sign-up which is fiddly.

Fixes #1104.

### Common Content?

Yes (though the video is actually a CYF video not really a common one).

### DIFF
--- a/common-content/en/blocks/retro/index.md
+++ b/common-content/en/blocks/retro/index.md
@@ -12,14 +12,14 @@ hide_from_overview=true
 
 {{<tabs name="Run a retro">}}
 {{<tab name="Video">}}
-{{<youtube>}}https://www.youtube.com/watch?v=TeKHx6GY3Zw{{</youtube>}}
+{{<youtube>}}https://www.youtube.com/watch?v=3rCllHgDl7s{{</youtube>}}
 {{</tab>}}
 {{<tab name="Text">}}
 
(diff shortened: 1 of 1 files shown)


SallyMcGrath (approved):
Fab!

PR #1183, opened 2024-11-07
Use site not $.Site

illicitonion:
## What does this change?

Use `site` not `$.Site`

These are equivalent when they happen to be called in a context that has `.Site`, but `$.Site` breaks if the current context doesn't have a `.Site`.

Because `site` _always_ works, and when they both work they're exactly equivalent, prefer to use it. This makes future refactorings less brittle.


SallyMcGrath (approved):
Oh this would make things so much easier! TIL! (Not sure how I missed this...)

PR #1187, opened 2024-11-09
Introduce taxonomy equivalent of menu map

illicitonion:
## What does this change?

Introduce taxonomy equivalent of menu map

This produces the same view (sharing code to do so), but allows using a taxonomy instead of a list of menus. The benefit here is a taxonomy gives you a place (the taxonomy term page) to define metadata that may be useful in presenting each term (e.g. attaching a description to each term).

The equivalences between menu and taxonomy maps are:

| Concept                                                       | Menu concept     | Taxonomy concept      |
|---------------------------------------------------------------|------------------|-----------------------|
| A category (e.g. a column to display in the default map view) | A menu           | A term                |
| A value in the category (e.g. a a card to show in the column) | A page in a menu | A value within a term |

There are some interesting compatibility questions in here which I will comment on in the PR I'm raising for this.

Note: This currently doesn't actually define any interesting metadata (other than weight) for each term in the tracks taxonomy - I will send this through as a follow-up PR, to separate out "the idea of a taxonomy map" from "the details of this particular taxonomy".

### Common Theme?

See description above, and line comments for specific considerations.

### Org Content?

Updates org-cyf-tracks to use a taxonomy not a menu-map.

## Checklist

cc @CodeYourFuture/dpg-team - I've left comments with some compatibility questions that seem relevant to the releases discussions.


daslerr:
@illicitonion I don't see any comments in the PR other than the inline comments in the code, and I didn't see any specific questions in those. In general, this all seems reasonable to me. Seems like a reasonable change, and including some backwards compatibility with a warning also seems reasonable to me. Is there something more specific I should be giving feedback on? 
I should say that for the draft of the automated release note, we adapted a GitHub template for repos using semantic versioning that essentially equates "breaking changes" with major version and any other changes were minor versions. That also seems reasonable to me, but I'm happy to adapt to whatever suits CYF's philosophy. That's possibly a discussion that's tangential to this PR, though.

illicitonion:
> @illicitonion I don't see any comments in the PR other than the inline comments in the code, and I didn't see any specific questions in those. In general, this all seems reasonable to me. Seems like a reasonable change, and including some backwards compatibility with a warning also seems reasonable to me. Is there something more specific I should be giving feedback on?

A reasonable question!

I think my specific questions are:
1. At some point we want to remove support for the legacy `map` frontmatter. At what point do you think that should be? Which I guess ends up being the question: How many minor releases should we wait, or how much time should we wait, between making a backwards-compatible shim and turning it into a breaking change?
2. What do we want to consider as a breaking change? It feels like stopping calling any existing partial, renaming any partial, and changing the expected params of any partial are all definitely breaking changes?
3. When we remove support for the `map` frontmatter, do we want to add an explicit `{{ if .Params.map }} {{ errorf "map is unsupported" }` branch, or do we want to just ignore any handling for it and fall back to whatever the layout would be even if map had never been a special thing?
4. More broadly, do we want to explicitly require opting in to specific layout types to avoid this problem in the future (i.e. instead of `if map { render map } else { render timeline}` do we want `if map { render map } else if timeline { render timeline } else { errorf "all pages must choose what kind of layout they are" }`?) This would require _every_ page to opt in to how it wants to be rendered (rather than some pages doing so and others falling back to a default). I think @SallyMcGrath has currently been modelling the `map` view as "just some frontmatter you can enable" (which is why it's documented in https://common.codeyourfuture.io/common-theme/front-matter/#map), but I think I model it as "you're changing the entire layout of a page, you just happen to be doing so via some frontmatter"... If we do the `else { errorf` way, removing an unsupported type automatically becomes a breaking change at build time, rather than defaulting to falling back to something else (and us needing to handle it specially if we want to make it actually break).

daslerr:
Those are indeed more specific. :) 

So first, let me say that from a community perspective, I'm of the opinion that it's much more important to decide on a release cadence and communicate it well than whatever the actual cadence is. I haven't been around CYF long enough to know what you guys would consider a reasonable schedule given your staff and resource constraints, so I don't think that's a determination I can make. Also, with my former PM hat on, I feel like this is straying into product decision territory. I think it would be best to loop in @kfklein15 on this discussion. 

My particular (by no means definitive) opinions on your questions:

1. My vague answer to "how long should we wait" is "a reasonable time for your core users to make the switch". In other contexts, I would maybe look at usage data to figure this out, but I'm not sure that's relevant here. For the users you expect to have, is it maybe more relevant to say "we'll continue to support the old way for a year [or some amount of time rather than number of minor releases]"? 
2. Sure, those all seem like breaking changes to me, but I may not be the best person to determine that. 
3. I don't think you'd need to leave explicit errors for legacy things that are unsupported. I think the documentation and the release notes should indicate that you're sunsetting a particular feature/function/whatever and when that will happen, and I think it's reasonable to include the warnings you've already included for the transition period, but when the specified time to sunset has come, it's just over and done. At this point, my understanding is that only one other organization is using common, so I think best case it's more likely that you'd have new people popping in to use the theme as it currently stands than you would have people looking for features they used to be able to use. 
4. This is a bit in the weeds from a development perspective for my skillset, I'll admit. But my gut feeling is that it doesn't seem necessary to opt-in to layout types specifically. But I'm happy to let other people have that fight. 

I hope it's a bit helpful to get a relative outsider's opinion, but again, I don't think I'm the right person to be dictating these decisions. I'm happy to document and support whatever direction CYF would like to go with a release cadence. 

I don't think the answers to these questions should hold up merging a PR, personally, but I'll wait for you to confirm that this is a satisfactory "review" before I approve it. I don't want to automerge anything you guys aren't ready for.

illicitonion:
I think that all makes sense, thanks @daslerr!

I'll leave for @SallyMcGrath to review and we can merge this - thanks for talking through this as a case study with me!

SallyMcGrath:
Great convo here both of you, thanks. 

Breaking change probably means you can't build your site when the only change you have made is updating the version from us. That's the same things said from the other side. 

In a call with Reshama more may be forthcoming.

reshamas:
@illicitonion 
We added a new label `Breaking Change` to this PR.

reshamas:
- Proposal: 2-cycle deprecation 
- Breaking Changes: add that label to whatever PRs may pose a breaking change for users in building the site

illicitonion:
Thanks for the discussion all!

> @illicitonion We added a new label `Breaking Change` to this PR.

Interesting... I put in some work to make this _not_ currently a breaking change - is the idea that we're collecting "things which introduce deprecations" as breaking changes? Or should I remove those back-compat affordances and make this a breaking change? Or should I remove the label, and add it in two cycles when we remove the deprecated back-compat code?

> Proposal: 2-cycle deprecation

Works for me! Just to be clear, does that mean:

Version n: Introduce deprecation
Version n+1: Deprecation stays
Version n+2: Add breaking change removing deprecation

?

> Breaking change probably means you can't build your site when the only change you have made is updating the version from us.

So in this example, if I remove the special-casing for looking for the `map` frontmatter, the site will continue to build, but it will be broken (any `map` pages will fall back to timeline layout rather than map layout). Do we consider this "breaking"?

And if we do, are we effectively saying our policy is: You should upgrade versions one at a time and address deprecation warnings as they come up. If you jump versions, things may break in ways you're not expecting without warning.

SallyMcGrath (approved):
I've spun it up and it looks great

illicitonion (commented):
Oh ugh! Apparently GitHub left my line comments as pending - I'm sorry @daslerr, I understand why you were confused about them now! Here they are for posterity!

illicitonion (commented):
I'm removing Breaking Changes from this because it currently isn't - happy to discuss changes to how we do this in the future!

daslerr:
> Oh ugh! Apparently GitHub left my line comments as pending - I'm sorry @daslerr, I understand why you were confused about them now! Here they are for posterity!

Ah, yeah, that makes a lot more sense now. :) Thanks!

PR #1194, opened 2024-11-12
Remove obsolete prereqs from Piscine

illicitonion:
## What does this change?

See discussion from https://codeyourfuture.slack.com/archives/C012UUW69S8/p1729604686501789?thread_ts=1729514033.295269&cid=C012UUW69S8

We will re-add error handling when we've re-added this to ITP.

The Jest updates are reflecting what we actually teach in ITP - the other things aren't super necessary they're Just More APIs.

Switch isn't needed and can be picked up as needed through Google.

### Common Content?

Yes, but this is maybe actually super CYF specific and maybe all the Piscine stuff should get moved into the Piscine module.


SallyMcGrath (approved):
With the caveat we will def return try/catch/throw next iteration

PR #1209, opened 2024-11-25
SDC: Merge How Computers Work and Tools into one module

illicitonion:
## What does this change?

Merges How Computers Work and Tools into one module.

I was about to create coursework repos for the first shell tools exercises, and the naming and chunking didn't super make sense to me. I didn't want to add a bunch of shell tools exercises to a module named "How computers work" when we also _have_ a "Tools" module.

I also think we're going to slightly de-emphasise the "How computers work" piece a bit, and have it be more part of learning more practical things.

Also moves `tr` from sprint 1 to sprint 2.


SallyMcGrath (approved):
Yes, I can't see an immediate problem - we can always review again if you need to back out.

PR #1210, opened 2024-11-28
Replace toBe with toEqual

illicitonion:
## What does this change?

Replace `toBe` with `toEqual`

These were meant to be changed before, but apparently I missed a few.

toEqual is a good default, and toBe is something that generally only makes sense in more limited use-cases.

### Common Content?

Yes, but only trivially


SallyMcGrath (approved):
I will also now remove toBe from my solutions and prep notes which I faithfully included just where specified

PR #1230, opened 2024-12-06
Share non-caching config across sites

illicitonion:
## What does this change?

Previously some sites had config and others didn't, and none of them
disabled getremote caching for production.

This meant that things like PR lists may be stale, subject to hugo's
default getremote caching. It also meant that changing this was a bit
whack-a-mole across the sites.

This changes every site to symlink in the same config (so we can
centrally edit all sites), to disable caching for getremote in
production, and to actually use the production config when deploying.

It also adds tests to make sure that every site is set up with similar baselines of config and deploy scripts.

### Org Content?

Updates config and build for all our sites.


SallyMcGrath:
Thanks for this @illicitonion 

My question is, instead of symlinking is there a way we can inherit the config via the module mount system? Only because it's preferable to use one system instead of two if possible. I actually don't know the answer to this question, it's not rhetorical.

SallyMcGrath (approved):

PR #1264, opened 2024-12-23
Two tiny tweaks: legends in flow, animated details

SallyMcGrath:
1. This is how you put legends in flow so it stays where it is and pushes siblings down instead of growing up like the default behaviour.  It is a bit esoteric if you don't Just Know From Life. 

2. While I am here, here's an animation on details that doesn't push content down in flow, so, the opposite.


illicitonion (approved):
Thank you!

Issue #1052, opened 2024-10-09
{{<our-name>}} doesn't work inside codefences


illicitonion:
On https://programming.codeyourfuture.io/structuring-data/sprints/3/prep/#setup we have a `<span>` literal in our code block:

<img width="935" alt="Screenshot 2024-10-09 at 10 59 10" src="https://github.com/user-attachments/assets/580bd603-cf6e-4fc7-b02d-6272361f5e10">

illicitonion:
It looks like this only takes effect inside tabs - if we're not in a tab things Just Work properly.

illicitonion:
Tabs got reworked! 🎉

Issue #847, opened 2024-08-02
Retro block uses link to Google timer not a timer shortcode


illicitonion:
e.g. in https://curriculum.codeyourfuture.io/js1/sprints/1/day-plan/#retro in the "Text" tab

Tim-Quattrochi:
Hi, can I take on this issue?

illicitonion:
Hi @Tim-Quattrochi! That would be great if you would be interested! I think the content you'd want to edit is in https://github.com/CodeYourFuture/curriculum/blob/main/common-content/en/blocks/retro/index.md and you can see an example of the timer shortcode in https://github.com/CodeYourFuture/curriculum/pull/846/files#diff-4a82407713fcc1fa17da022bc461ffe66932e864dc7176b4c78ee7fe98334f29

Thanks!

SallyMcGrath:
👀 That looks to me like a Hugo markdown bug. It's getting confused and parsing the html as markdown commands

> Like a blockquote

I will take a look and see if there's something easy we can do to jump this out of the render loop

SallyMcGrath:
Update on this: it's specifically in tabs and it's caused by this line https://github.com/CodeYourFuture/curriculum/blob/e8be388b340d5540e51719f3ef48173ece5d0ad4/common-theme/layouts/shortcodes/tabs.html#L27

The render string is rendering the timer (it thinks the > is a blockquote) before Hugo can take care of it. 

Tabs are always such a pain!

Tim-Quattrochi:
> Update on this: it's specifically in tabs and it's caused by this line
> 
> https://github.com/CodeYourFuture/curriculum/blob/e8be388b340d5540e51719f3ef48173ece5d0ad4/common-theme/layouts/shortcodes/tabs.html#L27
> 
> The render string is rendering the timer (it thinks the > is a blockquote) before Hugo can take care of it.
> 
> Tabs are always such a pain!

Nice research. How did you want to take care of this? Is this something where the tabs shortcode needs to be modified or am I rendering wrong?

SallyMcGrath:
Oh you've done nowt wrong. The opposite! You've found a genuine bug.  I think it will need the tab partial to be rewritten so they don't render. Reading the file it looks like the tabs were from Docsy and really I should have just written them myself.  But I haven't got time to do that now as we're launching all these courses. 

[...]

PR #1268, opened 2024-12-28
1068 Tabs Have Got To Go

### LINKED ISSUE #1068: Tabs Have Got To Go
**Describe the bug**

Our tabs are fragile and terrible and break in so many ways. Let us count the ways!

1. The double markdown render loop breaks some rendering in unpredictable ways #954 #76 #245 #288 #145 ...
2. The html is fine but we probably should move to inert now it's in baseline
3. The web component might not be necessary any more
4. It's possible to write broken tabs #1067 
5. The markup is cumbersome and I always forget it
6. Add your hatred and grievances here if there's room after all mine

☝️ 
**How could this be fixed?**

Write new tabs. We foolishly re-used Kubernetes/Docsy tabs and this is the source of our pain.

Replace the old markup too maybe

**Link to content**

**Additional context**

SallyMcGrath:
## What does this change?

Removes old tabs and replaces with a simpler format that is less prone to breaking its inner shortcodes. 

Why does it work better? Ohhhh, well, it just reduces the number of renders it has to go through which could screw up and escape html into the view, basically. And it's a bit easier to type as pressing the shift key is hurting my hands these days.

Why reduce renders? What does this mean? How does Hugo render content? Like most template engines, it more or less renders from the innermost piece and works its way outwards. Hooks, shortcodes, partials, layouts, base. At any point you might `| markdownify` a string. If you do nothing, all the strings will accumulate into `.Content` on the Page and be rendered there. Hugo will try to guess what is markdownifiable at this point. Typically then it gets some things wrong and we humans fix this by marking strings to be explicitly markdownified before they blobulate into .Content. 

https://gohugo.io/functions/transform/markdownify/

But where this goes wrong is twofold:

1. Markdownify only processes the markdown in front of it. It can't process the actual Page context, so if your markdown accesses the page context, for example to make a link or call an image, it could go wrong. 
2. HTML escaping -- markdownify cannot cope with whitespace, because markdown has, stupidly, whitespace formatting tricks (which should be illegal). So any time you are markdownifying you had better also rigorously chomp your whitespace in any HTML templating. In Hugo you do this with a dash`{{-` 

You can stop here and be happy with your life. Or you can read on and fill in your Hugo model a bit and understand why this ranges all over so many files when it claims to just be about tabs.

#### 67 files she changed, 47 markdown

So why are we using markdownify if it's so rubbish? Well, maybe we shouldn't any more. It's just… how it used to work. It used to be the only way to do this. It's also explicit -- it's easier to see that we want whatever is in this variable or dot to probably format as text or code or links etc.

https://gohugo.io/functions/transform/

So what's the alternative to markdownify for partials that contain markdown?

https://gohugo.io/methods/page/renderstring/

`RenderString` is now what `markdownify` does under the hood nowadays. It used to work quite differently to `markdownify.`  (You can also pass different rendering options like pandoc but let's not get distracted.) Whenever you get problems with your `markdownify`, someone will suggest using `$.Page.RenderString` instead. It's used like `z-index:1000` in CSS - like magic beans. But this isn't helpful often because

> It requires a `Page` object ! (And that's where this falls down)

It's clearly a method on `Page`. It simply doesn't apply to rendering outside this context. It will fail out on most of the places we transform md as those aren't Pages. (Like, all remote blocks). But then how does `markdownify` work if I'm saying it's `RenderString` secretly anyway? Well, it sets the page context to the site home page and pretends.  

Sooo we can use `page`. For many contexts we can and should use `page.RenderString`. We should use it whenever we are operating on normal partials rendering local markdown content. Even with the block system, since 0.111.0 we can use lowercase `page` to access "the current page" even if that's out of scope! 

So I've done that

https://github.com/CodeYourFuture/curriculum/pull/1268/files?file-filters%5B%5D=.html&show-deleted-files=true&show-viewed-files=true

https://gohugo.io/functions/global/page/

But notice, scroll down, we _shouldn't_ use this with shortcodes or we might end up with cached rendered shortcodes displaying on other pages with the wrong insides. And shortcodes are where all our pain resides. 

#### Yonder magic beans are ye

While we rest here, quiet and sad, another magic bean is the `{{% render first syntax %}}`. Typically we write shortcodes with angle brackets `{{<shortcode>}}` . If we use the `%` then the shortcode is rendered and passed to goldmark as html. Another way to say this is `<` means "contains only html" and `% `means "contains markdown and html". Even though in practice it's fine to pass html into `<` because you are dealing with that in your shortcode. This is almost never relevant to your problem but will be suggested frequently. 

#### SafeHTML, she cried

Another magic bean, applied randomly, is `safeHTML`. Does nowt in almost all of these cases because `markdownify`, `RenderString` and `.Content` already return `safeHTML`. It's really meant for when you are constructing a string with, say, printf and then you want that parsed as html subsequently. 

https://gohugo.io/functions/safe/html/

#### Whitespace crisp and even

Weirdly, the most unloved magic bean – chomping your whitespace – is nearly always the right answer for us. `{{-`. This is because goldmark will interpret whitespace as either a paragraph, blockquote, or code block. It will then close out your parent shortcode and dump your child shortcode inside that phrasing block, which will either look broken or actually be broken. This is the main risk of transforming at the component level and the main solution is to clean up your whitespace. This could include moving comments to the bottom of a file. It could also include passing RenderString the opt "inline".

So, you might wonder at this point why we are transforming per partial instead of bringing everything up to a page and then rendering on there. I wondered this too, and I tried it, and everything broke. I felt so weary at this point that I could go no further. 

#### Tread thou in them boldly

You can explore how this works by removing the markdownify transform from note.html. On a day plan, say,  the notes coming from a readme block will be transformed, but the local notes will not. This is because the readme notes are being transformed anyway by the markdownification in readme.html. There's no markdownify in local.html, so the note comes through untransformed. So, should local be markdownified?  If we write a note shortcode in a page, just any content page, it will be rendered via .Content. And in fact all the shortcodes are easily and beautifully rendered inside each other at this level inside Page. It all Just Works. But we do want to compose pages from multiple sources including remote sources, so here we are. 

I updated the `markdownify` calls outside of shortcodes to either `.Page` or `page .RenderString`s and went forth through the rude winds' wild lament.

And the bitter weather.

#### Render hooks and bring me wine

But this is Hugo, so what is the confounding exception? Oh, that's render hooks! Render hooks hook IN to the render and extract pieces from it, to be processed separately. That's how syntax highlighting works in code fences, but you can hijack any segment of markdown and process it first, then pass it up through all your contexts as safeHTML. Render hooks are almost a better version of shortcodes. It feels like they are. I prefer them in most cases as they don't break so often or as weirdly. But of course they do break. If you pass a shortcode into a render hook by using, for example, a GFM note and putting a youtube video inside it, then your hook might simply give up, close itself out, and dump the nested shortcodes underneath. So really render hooks should be terminal components only? GFM Notes don't really fit in this paradigm, but they exist, so…. Hahahahugo!

Merry Christmas!

### Common Theme?

Fixes #1068
Addresses #847

### Org Content?

Module | Sprint | Page Type | Block Type

## Checklist


SallyMcGrath:
BTW this is "breaking" the deploys because this PR is being interpreted as unclosed shortcodes in the docs, sigh! Let's just merge it and it will heal itself.

common-content/en/module/js1/comparison/index.md
```diff
@@ -28,13 +28,11 @@ Whenever we call `formatAs12HourClock` we expect it to return a particular value
 However, the **current output** is what the function _actually_ returns right now.
 
 👉🏼 [Work through the exercises](#current-1)
-
-{{</tab>}}
-{{<tab name="Exercise 1">}}
+===[[Exercise 1]]
```
illicitonion, line 31: This is looking pretty broken? https://deploy-preview-1268--cyf-programming.netlify.app/structuring-data/sprints/2/prep/#comparison

Screenshot 2024-12-30 at 11 30 07

common-theme/layouts/shortcodes/tabs.html
```diff
@@ -1,32 +1,59 @@
...
-        <div id="{{ $id }}" class="c-tabs__panel" role="tabpanel" aria-labelledby='tab-label-{{ $id }}'>
-      {{ end }}
-        {{- with .content -}}
-          {{- . | $.Page.RenderString -}}
-        {{- end -}}
-      </div>
+    {{- range $index, $tab := $tabs -}}
+      {{- if ne $tab "" -}}
+        {{- $id := printf "%s-%d" $tab_set_id $index -}}
+        {{- $tabParts := split $tab "]]" -}}
```
illicitonion, line 47: I'd be tempted to make this a slightly more distinctive closing delimiter, if our method of parsing is just splitting on it, I can imagine having `]]` in some text somewhere (e.g. if ever having an `if [[ foo ]]` in a shell script)... Maybe something like:

```suggestion
        {{- $tabParts := split $tab "]]===" -}}
```

illicitonion (changes requested):
Looks great in theory! Seeing a few issues in practice...

common-content/en/module/js1/comparison/index.md
SallyMcGrath: Oh that was my working test block, dammit, so it still has old code in. Sigh! Thanks!

SallyMcGrath: No I've broken it somehow in the night! Just by sleeping! This is why sleep is for fools.

illicitonion: What about this one? :D

https://deploy-preview-1268--cyf-programming.netlify.app/structuring-data/sprints/3/prep/

Screenshot 2025-01-02 at 09 33 02

illicitonion:
LGTM, ship it!!

One thing that is still broken (but were broken before) that I thought this was going to fix: https://github.com/CodeYourFuture/curriculum/issues/1052

https://deploy-preview-1268--cyf-programming.netlify.app/structuring-data/sprints/3/prep/#starting-a-project

Screenshot 2025-01-02 at 09 55 57

illicitonion (approved):

SallyMcGrath:
> LGTM, ship it!!
> 
> One thing that is still broken (but were broken before) that I thought this was going to fix: #1052
> 

That's definitely related so not unreasonable to think it- it's this central Hugo question of at which layer of the onion skin do we want this code to be rendered. In the case of ourname, we _always_ want it to process first and pass to any other parent as safeHTML. We never want it to be markdown. Inside tabs it's processing the code first and then this shortcode. (If we move ourname out of tabs it renders fine. ). 

Screenshot 2025-01-02 at 10 26 08

PR #1274, opened 2024-12-30
Ensure every site configures markup

illicitonion:
This avoids needing to notice and play whack-a-mole when things aren't
configured as we expect.

Also, add a --fix mode which will try to fix things for you.


common-docs/hugo.toml
```diff
@@ -63,20 +63,24 @@ maxAge="6h"
 getenv = ["^HUGO_CURRICULUM_GITHUB_BEARER_TOKEN$"]
 
 
+theme = "common-theme"
```
SallyMcGrath, line 66: I actually don't think we need to declare common theme any more  if we're importing the module btw

org-cyf-sdc/hugo.toml
```diff
@@ -24,17 +24,22 @@ baseURL = "https://sdc.codeyourfuture.io/"
 [params]
 googleFonts="https://fonts.googleapis.com/css2?family=Space+Grotesk:wght@300;500;700&family=Inter:wght@400;600&display=swap"
 
+#this makes the section nav go the right way
+# because of this 'unexpected behaviour' https://gohugo.io/methods/page/nextinsection/
```
SallyMcGrath, line 28: Can this one not be inherited from theme? 

Actually maybe this should be tidied up a bit more. What can we inherit from org-cyf-theme module?

https://gohugo.io/getting-started/configuration/#configure-page

SallyMcGrath (changes requested):
Thank you for this. I do want this - but before we require it, can we check what can be inherited?

I'm not sure we do need to copy pasta as much config as we currently do, so would like to not bake that in (except the necessary). Can you read over this and have a think?

https://gohugo.io/getting-started/configuration/#configure-page

PR #1277, opened 2024-12-30
Allow filtering PRs by GitHub reviewer

illicitonion:


org-cyf-itp/layouts/_default/needs-review.html
```diff
@@ -9,6 +9,8 @@
 
   <div id="pr-list">Loading...</div>
 
+  <input type="text" id="reviewer-filter" placeholder="Filter by reviewer GitHub username" />
```
SallyMcGrath, line 12: Placeholders are not a reliable or accessible way of labelling an input.
```suggestion
  <label>Filter by reviewer <input type="text" id="reviewer-filter" placeholder="Input a GitHub name" /></label>
```

SallyMcGrath (changes requested):
TY! Please place this input _before_ your templates. This is because at the moment a screenreader will not be able to perceive the effect of the filter as it applies before their current position in the document. They drive forwards/down the page, linearly. 

For extra points, include a polite live region which carries the message "n results". Place this directly after the labelled input, so a screenreader does not have to read through an empty page to know that it is empty. Update the live region with the number of results and the entire message will be read into the announcer.

illicitonion:
How do you feel about `<select>`s?

SallyMcGrath:
> How do you feel about `<select>`s?

So long as the control has an associated label, I'm fine with it. I might draw your eye to a datalist, which you can see an implementation of here on line 52 https://github.com/CodeYourFuture/curriculum/pull/1279/files?file-filters%5B%5D=.html&show-viewed-files=true#diff-1c4fdbf6a7c343a49f24680d6b7f5637163a64a405fc0466a3c4adb8e850ccf8

gives more user interaction options like type to complete

org-cyf-itp/layouts/_default/needs-review.html
```diff
@@ -7,9 +7,9 @@
   {{ end }}
   <div id="overview" class="c-pr-overview"></div>
 
-  <div id="pr-list">Loading...</div>
+  <select id="reviewer-filter"></select>
```
SallyMcGrath, line 10: label it please! 🙏

SallyMcGrath (approved):

PR #1281, opened 2025-01-02
Fix: cleanup after tabs work

SallyMcGrath:
this one was missed somehow so is breaking data-groups/sprints/2/prep/#access-with-variables

Have searched and found no more

#1268


illicitonion (approved):
Weird 😂

PR #1287, opened 2025-01-06
Update issue links to point at correct module

illicitonion:


kfklein15 (approved):
Thanks for changing this so swiftly.

PR #1290, opened 2025-01-07
Update steps in ITD workshops

illicitonion:
## What does this change?

We changed the steps of the course, but not in the workshop.

### Org Content?

ITD workshops


AmyCYF (approved):
Looks good to me

PR #1294, opened 2025-01-09
Give friendly error on empty backlog repo

illicitonion:
Currently you get a complicated error about a nil dereference. This makes it clear what happened.


SallyMcGrath (approved):
I'm happy with this but tbh thought it did warn already? It comes up in yellow? But happy for this!

PR #1304, opened 2025-01-13
Fix YouTube embed

illicitonion:
## What does this change?

Fix YouTube embed

We don't chomp whitespace, so it ends up in the URL


SallyMcGrath (approved):
we probably should chomp whitespace - add it to the other things we should do eg clips

PR #1308, opened 2025-01-14
Clarify access to How Computers Really Work

illicitonion:
Matthew Justice has very kindly donated a PDF of the book to CYF for use by our trainees, which means we have a much better answer for "how do you get this book if you can't afford it?" now!

This block overrides the common-content version of this block, because it's a CYF-specific donation.


SallyMcGrath (approved):
this is a perfect use of the local overrides 🌟

PR #1313, opened 2025-01-20
SDC Decomposition Sprint 2

illicitonion:


illicitonion:
Merging after discussion with Sally - we're happy with the structure of this, and want to get it up so people can start looking at the SDC, but will probably make extensive copy edits :)

PR #1314, opened 2025-01-21
Add teaching strategies/guidelines

illicitonion:
We talk about these things a lot, but I don't think we have them written down anywhere. Now we do!


illicitonion:
Before `activities` was in `teaching/activities`. I wanted to add a `teaching`. If I added things to `teaching` directly, it meant `activities` became a sub-menu. So I pulled `content/teaching/activities` up to `content/activities` so it stays its own top-level menu.

SallyMcGrath (approved):
So long as this doesn't muck up the taxonomy, seems great!

PR #1295, opened 2025-01-10
SDC Decomposition Sprint 1

illicitonion:
## What does this change?

Adds the first Decomposition sprint.

It's a slightly unusual sprint. It has a _lot_ of prep content, and the only backlog is basically: "Follow the prep, and actually do all the things".

I'm a bit concerned at how much it's mixing "background reading" vs "active reading to do stuff", but also hoping this leads to the reading actually seeming relevant and clicking...

### Common Content?

Adds a bunch of blocks

Removes/renames a handful of blocks no one has ever used and which never had any content.

### Org Content?

Sets up some SDC sprints.

PR #1316, opened 2025-01-23
Feature: bare bones labeller

SallyMcGrath:
## What does this change?

I've just put together a really bare bones labeller. It is similar to the multiple choice component. In this one they drag and drop labels onto content to classify them. 

It seems like this is a key task in SDC and it's easy enough to make. Here's a sketch for you to see what you think @illicitonion @40thieves

Re #1295 

### Common Theme?
Yeah it's item-labeller.js and item-labeller.html

```

{{<label-items>}}
[LABEL=label 1] Something here some kind of content  
[LABEL=label 2] Some `more content` hurray  
[LABEL=label 2] Some more content hurray  
{{</label-items>}}
```


illicitonion (commented):
Ok, I've tried using this in anger - two pieces of feedback:

1. I think it'd be useful to put the whole component in a bordered-box of some kind. Right now I don't think it's super obvious where the component ends (e.g. that the feedback at the bottom is part of the component, and even which lines of text are things that should be labelled)
2. I think we should give a "whole exercise" piece of feedback, either as well as or instead of the latest-labelled-statement feedback. Currently if you correctly label one thing, there's no feedback saying "You got one right, but you have three still to do", which I think would be useful.

Also FYI this increases our minimum supported hugo version - I was previously using 0.136.2 locally, and these `strings` functions were added in 0.136.3. Not a problem, but just an FYI

(Also we're going to get broken at some point, I tried jumping all the way to latest, and apparently error handling has changed somewhere... But that's a problem for another day!)

illicitonion:
I tried putting a `<label-items>` inside a `<note>` to work around the border-box issue and ended up with a bunch of literal HTML getting rendered :)

SallyMcGrath:
> 
> Also FYI this increases our minimum supported hugo version - I was previously using 0.136.2 locally, and these `strings` functions were added in 0.136.3. Not a problem, but just an FYI


I just scarfed them all off this page https://gohugo.io/functions/strings/ and didn't investigate further. It did seem a bit easier than last time to clear up the trailing characters.

illicitonion (approved):
Thank you!

PR #1317, opened 2025-01-23
Tooltips allow different inline and title text

illicitonion:
## What does this change?

Tooltips allow different inline and title text

I frequently want to pluralise the inline text, but have the title of the tooltip be singular.

### Common Theme?

Yes - adds one extension point to tooltips. Backwards-compatible.


illicitonion:
I think this got merged elsewhere...

PR #1318, opened 2025-01-23
Remove the block concept of facilitation

illicitonion:
## What does this change?

Remove the block concept of facilitation

It's not used for anything.

### Common Content?

Removes property from all common content.

### Common Theme?

Removes from the block archtetype.

### Org Content?

Removes from a few blocks there too.


SallyMcGrath (approved):
TY!!

PR #1323, opened 2025-01-28
Fix up register

illicitonion:
* Rename course back to cyf-itp

This is hard-coded in the register handler.

Also add a check that courses have matching data.

* Symlink submission-created across courses 

Otherwise the Piscine doesn't get register events handled


SallyMcGrath (approved):
lgtm

PR #1325, opened 2025-01-28
~Final tweaks to Piscine

illicitonion:
* Note that groups may be 2 people.
* Add temporary block for talking through breaking down requirements. We hope to remove this in the future, but we would like our day plans to be accurate for the course we're delivering.
* Move team-selection to happen before morning break.

### DIFF
--- a/common-content/en/module/piscine/practice-break-down/index.md
+++ b/common-content/en/module/piscine/practice-break-down/index.md
@@ -1,0 +1,24 @@
++++
+title = "Practice breaking down a requirement"
+time = 60
+objectives = [
+  "Break down a requirement into approachable tasks.",
+  "Coordinate multiple team members in executing tasks in parallel.",
+]
+[build]
+  render = "never"
+  list = "local"
+  publishResources = false
++++
+
+In team projects, it's important that we break large tasks down into small tasks.
+
+It's also important that we can coordinate across our team. This requires having shared understanding of who will do what, and how the work we do will interact.
+
+We will practice this together on the first two requirements of the project.
+
+{{/*
+  TODO: The Piscine is meant to be pure evaluation with no teaching.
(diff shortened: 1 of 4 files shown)


common-content/en/module/piscine/practice-break-down/index.md
```diff
@@ -0,0 +1,24 @@
...
+  render = "never"
+  list = "local"
+  publishResources = false
++++
+
+In team projects, it's important that we break large tasks down into small tasks.
+
+It's also important that we can coordinate across our team. This requires having shared understanding of who will do what, and how the work we do will interact.
+
+We will practice this together on the first two requirements of the project.
```
40thieves, line 18: Did we want to document the broken down tickets here?

illicitonion: Works for me! I think you signed up to do the break-down, so shall we merge this and your can PR to it?

Also thinking about the break-down, it occurs to me that the first two tickets don't involve much cross-team coordination - do we maybe want to include in our break down something that e.g. requires agreeing on a serialisation format? Or would that be too much solving implementation things for them?

40thieves: Yeah, I realised that requirement 2 is most of the rendering aspect of the project, plus the de-serialisation.

It's actually surprisingly tricky - I think I've also come around to the conclusion that thinking about the serialisation is probably the best topic to tackle here.

40thieves: > so shall we merge this and your can PR to it?

Yep sounds good!

illicitonion: Sounds good :) will merge for now and look forward to your PR!

PR #1331, opened 2025-01-31
feature: issue shortcode

SallyMcGrath:
## What does this change?

This is so @40thieves can embed an issue in markdown

### Common Theme?
 /shortcodes/issue

Docs here /common-theme/shortcodes/issue/


illicitonion (approved):
Nice!

PR #1338, opened 2025-02-06
Markdownfiy labeller heading

illicitonion:


SallyMcGrath (approved):
I think this won't p wrap - if we find it p wrapping we will have to retrench

PR #1339, opened 2025-02-06
switch from marker to before

SallyMcGrath:
## What does this change?

This fixes a display bug in Safari. Safari does not implement ::marker completely.
https://bugs.webkit.org/show_bug.cgi?id=204163

So I removed ::marker and used ::before. 

I also removed the font size as the new emoji is the right size on OSX.

@illicitonion


illicitonion (approved):
Thank you!

PR #1355, opened 2025-02-17
Tweak logic day-plan timings

illicitonion:
* Move rats to morning
* Give more time for games - it's games week!
* Give more time for logic teaching
* Add in a regular demos slot - demos _every_ week!
* Switch back to retro - we're doing retros rather than wraps for most of SDC.


SallyMcGrath (approved):
Whoop!

PR #1366, opened 2025-02-26
Publish SDC PR dashboard

illicitonion:


SallyMcGrath (approved):
We should build this in as an optional template at some point

PR #1384, opened 2025-03-11
Give tabs horizontal scrolling

illicitonion:
Currently if a tab overflows it makes the page (potentially very much) wider. In-tab scrolling seems more friendly.


SallyMcGrath (approved):
no objection

PR #1401, opened 2025-03-29
Add new step 1 to the list

illicitonion:
This doesn't rename all the blocks, which we should probably do at some point. (But we should probably name them to be semantically named rather than numerically named).


kfklein15 (approved):
Checked the preview and everything looks good.

PR #1407, opened 2025-04-08
Re-add technical interview guide

illicitonion:
This was in the old syllabus and got a bit lost. This pulls in the old content pretty much verbatim and updates the markdown a bit.


SallyMcGrath (approved):
LGTM

PR #1413, opened 2025-04-14
Misc edits in prepration for the next ITP run

illicitonion:
See preview in https://deploy-preview-1413--cyf-programming.netlify.app/

These were discussed with the CYF staff guiding the first sprint in each region:
* Make the Welcome sprint energiser just be "Introduce yourself"
* Consistently label energisers as energisers, and invite folks to think of / bring their own.
* Call out that people should make name tags for the first few weeks.
* Add an (optional) call to hang out after class.
* Add a Code of Conduct block to the Welcome day-plan.
* Update that PRs must be complete before steps are submitted.
* Consistently suggest that Miro post-it workshops can be done on Miro or with post-its.
* Move Slack Libs from an energiser to a workshop-ish thing earlier in the course.
* Some copy edits on some activities
* Stop `{{<our-name>}}` from adding a trailing space


kfklein15 (approved):
Small but important changes. Thank you!

PR #1428, opened 2025-04-30
Edits to Complexity sprint 1

illicitonion:


common-content/en/module/complexity/big-o/index.md
```diff
@@ -142,7 +151,9 @@ Oh where have we seen this sequence of numbers before? ;)
 
 <summary>
 
-😱 **Exponential:** The runtime grows exponentially with the input size.</summary>
+😰 **Exponential:** The runtime grows exponentially with the input size.</summary>
```
SallyMcGrath, line 154:
```suggestion
😱 **Exponential:** The runtime grows exponentially with the input size.</summary>
```

common-content/en/module/complexity/big-o/index.md
```diff
@@ -163,8 +174,12 @@ You will explore this theory in your backlog, but you will find that you already
...
-- Trying every possible combination to unlock a password. Each extra character dramatically increases the possibilities. This is like naive recursion;  we'll talk about this more later.
+- Trying every possible combination to unlock a password. Each extra character dramatically increases the possibilities.
   {{< /label-items >}}
 
 > [!TIP]
 >  Big-O notation also describes space complexity (how memory use grows). Sometimes an algorithm's time complexity is different from its space complexity. We have focused on time here, but you'll meet space complexity analysis in the assigned reading. 
+
+Big-O notation is focused on the _trend_ of growth, not the exact growth.
+
+If we think about strings, one character may take up one byte or four. If we think about doubling the length of the string, we don't worry about what characters are actually in the string. We think about the _trend_ - the string will take _about_ twice as much space. If the string only contained four-byte characters, and we only added one-byte characters, the string is still growing linearly, even though it may not take exactly double the space.
```
SallyMcGrath, line 185: I think this might belong in the coursework ?

The trend is explained in the linked texts btw!

common-content/en/module/complexity/caching/index.md
```diff
@@ -1,11 +1,11 @@
 +++
 title = "Caching"
+time = 15
```
SallyMcGrath, line 3: Is this why the time is wrong on the prep?? The order?

common-content/en/module/complexity/memory-consumption/index.md
```diff
@@ -34,11 +34,14 @@ const userRoles = ["Admin", "Editor", "Viewer"]; //An array of 3 short strings
 const userProfiles = [ {id: 1, name: "Farzaneh", role: "Admin", preferences: {...}}, {id: 2, name: "Cuneyt", role: "Editor", preferences: {...}} ]; // An array of 2 complex objects
 ```
 
-Different kinds of data have different memory footprints:
+Different kinds of data have different memory footprints. All data if fundamentally stored as bytes. We can form intuition for how much memory a piece of data takes:
```
SallyMcGrath, line 37:
```suggestion
Different kinds of data have different memory footprints. All data is fundamentally stored as bytes. We can form intuition for how much memory a piece of data takes:
```

common-content/en/module/complexity/memory-consumption/index.md
```diff
@@ -34,11 +34,14 @@ const userRoles = ["Admin", "Editor", "Viewer"]; //An array of 3 short strings
...
 
-- Numbers or booleans use less memory than objects
+- Numbers are typically stored as 8 bytes. In some languages, you can define numbers which take up less space (but can store a smaller range of values).
+- Each character in an ASCII string takes 1 byte. More complex characters may take more bytes. The biggest characters take up to 4 bytes.
 - The longer a string, the more bytes it consumes.
-- Objects and arrays need memory for their internal organisation _as well_ as the data itself.
+- Objects and arrays are stored in different ways in different languages. But they need to store _at least_ the information contained within them.
+  - This means an array of 5 elements will use _at least_ as much memory as the 5 elements would on their own.
+  - And objects will use _at least_ as much memory as all of the _values_ inside the object (and in some languages, all of the keys as well).
```
SallyMcGrath, line 44: Again, is this a bit too much info? We don't want to overload them.

common-content/en/module/complexity/n+1/index.md
```diff
@@ -55,11 +62,8 @@ The real `/home `endpoint avoids these problems by using efficient strategies:
...
-  feedback="No, but flip this and try again? | Right! That's a clear description. | No, this is so vague it describes nothing. | No, it's not about network endpoints."
-  correct="1">}}
\ No newline at end of file
+* Batching may reduce our responsiveness. It will probably take longer to fetch three users' blooms than one user's blooms. If we'd just asked for one user's blooms, then the next user's, we probably probably could've showed the user _some_ results sooner. Batching forces us to wait for _all_ of the results before we show anything.
+* Caching may result in stale results. If we store a user's most recent blooms, and when they re-visit the page we don't ask the database for the most recent blooms, it's possible we'll return old results missing the newest blooms.
+* Pagination means the user doesn't have complete results up-front. If they want to ask a question like "has this user ever bloomed the word cheese", they may need to keep scrolling and searching to find the answer (with each scroll requiring a separate network call and database lookup).
```
SallyMcGrath, line 69: I'm just loving the sentence "has this user ever bloomed the word cheese" !

common-content/en/module/complexity/worked-example-duplicate-encoder/index.md
```diff
@@ -0,0 +1,119 @@
...
+    "Identify that speed isn't the only factor in choosing the best code.",
+]
++++
+
+Let's consider two possible approaches to [the Duplicate Encoder Codewars exercise](https://www.codewars.com/kata/54b42f9314d9229fd6000d9c/train/javascript).
+
+The problem can be stated as:
+
+{{<note type="exercise">}}
+The goal of this exercise is to convert a string to a new string where each character in the new string is `(` if that character appears only once in the original string, or `)` if that character appears more than once in the original string. Ignore capitalization when determining if a character is a duplicate.
```
SallyMcGrath, line 21: This sentence is insanely long
```suggestion
The goal of this exercise is to convert a string to a new string where, for each character:

1. if the original character appears only once, replace with `(``
1. if the original character appears more than once, replace with `)` 

Ignore capitalisation when deciding if a character is a duplicate.
```

SallyMcGrath (changes requested):
Lovely! A few small text changes.

common-content/en/module/complexity/caching/index.md
illicitonion: Yeah, before the `time` properties were getting treated as `build.time` (because they were nested in that section) rather than top-level `time` properties.

(I am such a fan of build errors when you set a property that isn't used...!)

common-content/en/module/complexity/memory-consumption/index.md
illicitonion: Maybe... Let's maybe have a chat about it?

I changed it because the previous content is kind of misleading - in most compiled programming languages, arrays and objects take exactly as much space as their contents/members/fields, so I wanted to avoid instilling an intuition that they have overhead where they don't necessarily...

common-content/en/module/complexity/worked-example-duplicate-encoder/index.md
illicitonion: Yeah, this was me literally copy+pasting the text from Codewars - I've just rewritten to not reference Codewars :)

common-content/en/module/complexity/big-o/index.md
illicitonion: I think the reading is quite vague on this - I think the only coverage is "Big O notation is used to express the worst-case scenario of an algorithm's runtime performance, abstracting away constants and lower-order terms to focus only on the most significant factors that affect the running time." which is easy to miss or mis-understand, so figured it was worth re-stating. Rephrased repetition is helpful...

common-content/en/module/complexity/memory-consumption/index.md
SallyMcGrath: Hmm, it's a good point.  I was trying to steer away from the python size of debacle I've encountered before with trainees and have oversteered.

common-content/en/module/complexity/big-o/index.md
```diff
@@ -108,9 +115,11 @@ line [100, 400, 900, 1600, 2500, 3600, 4900, 6400, 8100, 10000]
 
 What does this mean? It means that the time is the square of the input size: n\*n.
 
+An example is finding which elements in an array are present more than once. For each element, we need to check every other element in the same array to see if they're equal. If we double the number of elements in the array, we _quadruple_ the number of checks we need to do.
```
SallyMcGrath, line 118: TBH I had intended these just as a visual

common-content/en/module/complexity/big-o/index.md
SallyMcGrath, line 185:
```suggestion
Think about strings: one character may take up one byte or four. If we double the length of the string, we don't check which characters are in the string. We just think about the **trend**. The string will take _about_ twice as much space. If the string only has four-byte characters, and we add one-byte characters, the string is _still_ growing linearly, even though it may not take _exactly_ double the space.
```

SallyMcGrath (approved):
my edit on this

common-content/en/module/complexity/big-o/index.md
SallyMcGrath: Ok, let's just simplify this a bit then? I made a tiny edit

Issue #996, opened 2024-09-25
Horrific FOUC on Firefox 


SallyMcGrath:
**Describe the bug**

Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:130.0) Gecko/20100101 Firefox/130.0

Given a page on the curriculum
When I load the page
Then there's a jarring FOUC

**What is wrong with this content?**

Ugly I hate it!

**How could this be fixed?**

Make it go!

**Link to content**

**Additional context**

On a cursory glance, looks like I've reduced the fade to 10ms at some point for unfathomable reasons (maybe just fat fingered). This could be fixed by bumping this back up to 100ms. I think 200ms would be nicer. 

https://github.com/CodeYourFuture/curriculum/blob/874907ac6f636e00e745ae1b79c6c429957e0eef/common-theme/layouts/partials/head.html#L28

**Who else might need to know about this?**

Issue #966, opened 2024-09-20
Check on reliability of Python Tutor


illicitonion:
https://github.com/CodeYourFuture/curriculum/blob/main/common-content/en/module/js1/playing-computer/index.md embeds an iframe from pythontutor.com which is currently reliably 504ing.

We should make sure pythontutor.com is still intended to work (and comes back up), and if not, find a replacement.

FacuMa02:
Hi! I would like to collaborate in this issue.

illicitonion:
Amazing, thanks @FacuMa02!

The main action item here is to see if the site is back up, and see if the visualiser on https://programming.codeyourfuture.io/structuring-data/sprints/1/prep/#playing-computer is working properly. If everything looks good there, we can close this issue and know we're happy. If the site is down, or the widget doesn't work any more, we can talk about what next steps we may need to take.

Thanks for your help!

illicitonion:
It looks like Python Tutor is back and has been for a while - probably just a one-off blip :)

PR #1456, opened 2025-05-20
Corrected bug #966 by changing animation to 200ms

ajay0120:
### Common Content?

- [ no] Block/s

### Common Theme?

-  Yes

Issue number: #996 

### Org Content?

It changes the fade in time in animation from 10ms to 200ms as per the request of SallyMcGrath

## Checklist

@SallyMcGrath

### DIFF
--- a/common-theme/layouts/partials/head.html
+++ b/common-theme/layouts/partials/head.html
@@ -25,7 +25,7 @@
       margin: 0;
       padding: 0;
       height: 100%;
-      animation: fade-in 300ms;
+      animation: fade-in 200ms;
     }
     @keyframes fade-in {
       0% {


illicitonion:
Hi @ajay0120 - thanks for putting this together.

Looking at the change, it looks like this problem may already have been fixed? #996 has #1014 attached to it which changed this value from 10 to 300. Your description says you're increasing it from 10 to 200, but it looks like you're actually decreasing it from 300 to 200?

Did you reproduce the problem before trying to fix it? If so, maybe you can share a short video of the problem reproducing?

ajay0120:
Hi @illicitonion I have actually seen the bug #996 as open so I tried to resolve. But yeah as you said someone has changed it to 300 but as per bug report they have requested it to be 200ms so that's the reason I have raised a pr.

illicitonion:
Thanks for taking the time to try to help, but I think unless you can reproduce the problem at the moment we should probably close this PR :)

PR #1482, opened 2025-06-20
Improve some learning objectives

illicitonion:
LOs should always be testable statements ("Compare X and Y" rather "Awareness of"), and should ~never start "Understand" :)


kfklein15 (approved):
LGTM.
Hopefully I have managed to change the next LO correctly on my other changes.

PR #1486, opened 2025-06-25
Update some course metadata

illicitonion:
* Rename "Entry-Level Tracks" to "Post-ITP Tracks"
* Update course durations


kfklein15 (approved):
Well spotted!

PR #1487, opened 2025-06-27
Update tracks in menu

illicitonion:
* Remove "Coming Soon" completely - we now have an actual track
* Hide the SAP track from the menu - it's a WIP placeholder for now, and its presence is confusing/misleading.
* Rename "Guided study after ITP" to "Study after ITP" to fix the formatting a bit.


kfklein15 (approved):
Makes sense.

PR #1500, opened 2025-07-15
Update copy on welfare briefing

illicitonion:


org-cyf-itp/content/blocks/code-of-conduct/index.md
```diff
@@ -65,8 +65,26 @@ If in doubt, anyone can (and should) report things to any member of staff. And i
...
+<details>
+
+<summary>How to Share Your Concern</summary>
+
+1. **Write down what you observed or learned**, especially if it’s from a conversation. Try to use the other person’s words where possible.
+2. **Tell a CYF Staff Member**: They will then take appropriate action.
+   * We may not include you in the actions after reporting.
+   * If you're hesitant, you can first share your concerns without naming the affected individual.
+
+**Important Points to Remember**
```
chinar-amrutkar, line 83: Would it be worth adding something on the lines of the risk of a false positive is more acceptable than the trainee facing an issue and not being reported? Perhaps in simpler language

chinar-amrutkar (approved):

PR #1507, opened 2025-07-19
feat: add GitHub edit and issue buttons to curriculum pages (#552)

### LINKED ISSUE #552: Feature:  open an issue, edit on github buttons
Module(s): all
Week(s): all

On a previous version we had some nice edit on github buttons. We have them on the CYF+ site too. 

I want them to be pretty discreet and not get in the way, so probably hidden in some kind of flyout on the block and not made super obvious. 

Probably something like

`{{ with .File.Path }}<a href="https://github.com/etc etc{{ . }}">Fix me</a>{{ end }}` 

It would have to go on the block I guess because the page doesn't have the file paths of all the blocks.

We are quite good at encouraging trainees and vols to open tickets but having a little boopy button right there on the block would probably help. 

**Additional context**

We want to encourage more and more trainees to collaborate and co-create their learning journey. 

SHENIPRASHANTH:
Fixes #552

Added two helpful links at the bottom of each curriculum page:
1)"Edit this page on GitHub"
2)"Open an issue for this page"

This makes it easier for contributors to propose changes or report problems.

### DIFF
--- a/common-theme/layouts/_default/baseof.html
+++ b/common-theme/layouts/_default/baseof.html
@@ -2,7 +2,7 @@
 <!--[if lt IE 7]>      <html class="no-js lt-ie9 lt-ie8 lt-ie7"> <![endif]-->
 <!--[if IE 7]>         <html class="no-js lt-ie9 lt-ie8"> <![endif]-->
 <!--[if IE 8]>         <html class="no-js lt-ie9"> <![endif]-->
-<!--[if gt IE 8]>      <html class="no-js"> <!<![endif]-->
+<!--[if gt IE 8]>      <html class="no-js"> <![endif]-->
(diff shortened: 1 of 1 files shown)


illicitonion:
Hi! Thanks for putting this together!

Can you talk me through how you're testing this change locally? I tried checking out your branch an running the Intro To Programming site (by running: `cd org-cyf-itp && npm run start:dev`) and it gave me an error:

```
ERROR render of "taxonomy" failed: "/Users/dwh/src/github.com/CodeYourFuture/curriculum/common-theme/layouts/_default/baseof.html:23:82": execute of template failed: template: _default/taxonomy.html:23:82: executing "_default/taxonomy.html" at <.File.Path>: error calling Path: runtime error: invalid memory address or nil pointer dereference
```

I think this generally looks good, but I'm not sure it quite always works at the moment :)

PR #1511, opened 2025-07-21
Remove Online register option for in-person courses

illicitonion:


kfklein15 (approved):
Checked all pages. Online is kept for ITD and Launch only. All other pages have only the regions. Tracks have no region, and I think that is as I would expect.

PR #1523, opened 2025-07-29
New AI Step for ITD

Ara225:
## What does this change?

This adds a new AI step for ITD. This is currently only theoretical - there will be another pull request at some point to introduce another AI step after they have learnt HTML and CSS.
- Does the messaging toe the party line as it were? Would be interested to know if we have a responsible AI use policy I could link to
- The description is too complex for the curriculum platform - it will have to be in a Google doc

### Common Content?

- [ ] Block/s

### Common Theme?

- [ ] Yes

Issue number: N/A

### Org Content?
ITD block 2

## Checklist

### DIFF
--- a/org-cyf-itd/content/steps/two/index.md
+++ b/org-cyf-itd/content/steps/two/index.md
@@ -1,6 +1,6 @@
 +++
 title = 'Two'
-description = 'Make a Cover Letter with Generative AI'
+description = 'Learn how to use AI appropriately'
 layout = 'prep'
 emoji= '🤖'
 menu_level = ['module']


org-cyf-itd/content/blocks/step-2/instructions/index.md
```diff
@@ -12,87 +12,78 @@ time= 90
...
+A CYF student asks an AI tool to help them understand a complex CSS topic. The information the AI gives doesn't sound right. 
 
-Software for Lawyers Ltd makes case management software. Their software helps lawyers manage caseloads and track time spent. The development team is small and close-knit. They are looking to add a Junior Software Developer to their team.
+Which of the below might be appropriate responses? 
 
-###### Requirements
+1. Googling the topic and look for reliable sources to clarify the point.
+2. Trying to persuade the AI to check again.
+3. Accepting this as the truth because the AI seemed to know what it was talking about.
+4. Asking a volunteer whether the AI is right
```
illicitonion, line 45:
```suggestion
4. Asking a volunteer whether the AI is right.
```

org-cyf-itd/content/blocks/step-2/instructions/index.md
```diff
@@ -12,87 +12,78 @@ time= 90
...
-{{</note>}}
+Which of the below are NOT appropriate responses? 
 
+1. Asking volunteers for guidance at a session
+2. Giving an AI tool their assigned problems and ask it to solve them for them. They then copy and paste the output, making no edits.
+3. Asking an AI tool to explain the concepts they find difficult to understand.
+4. Attempting to solve the problems, then giving up asking an AI tool to fix their code. Then copying and pasting the answer without understanding the changes it made.
+5. Using Google to find articles on the topics they are struggling with. 
+6. Copying another student’s answers.
+7. Sending a direct message to a active volunteer
```
illicitonion, line 59: It'd be good to be clear that the problem with 2 wasn't just "making no edits" :)

```suggestion
7. Sending a direct message to a active volunteer
8. Giving an AI tool their assigned problems and ask it to solve them for them. They then copy and paste the output, making small edits.
```

Ara225:
Changes accepted and I think I found the bit you were referring to about the ghosts of the previous step. Can't see any other bits left

Additionally, I think that this will need some guidance added to the canvas on the ITD channels because the people we have reviewing ITD can be quite junior and may not think it through. I am thinking something like this ( I think it may be too patronizing):
## Question 1
*A CYF student is assigned coursework on a CYF course. The assignment is to build a homepage for a library which displays their twenty newest books. It requires that they use CSS Grid (a CSS feature used to display HTML elements in a grid format). They think that ChatGPT could help them complete this task.* 

*Which of the prompts below is an appropriate use of AI?* 

1. *Generate a webpage to display information about books in a grid format.* ❌
2. *Use CSS Grid to build a webpage to display twenty digital books* ❌
3. *Build a homepage for a library website* ❌
4. *I want to learn more about CSS Grid. Please explain the core concepts simply with examples.* ✔ - This is correct because it doesn't ask the AI to do the work - it simply asks it for help understanding a core concept. The others ask AI to do the work (mostly somewhat badly)

## Question 2
*A CYF student asks an AI tool to help them understand a complex CSS topic. The information the AI gives doesn't sound right.*

*Which of the below might be appropriate responses?*

1. *Googling the topic and look for reliable sources to clarify the point.* ✔
2. *Trying to persuade the AI to check again.* ❓ - This can sometimes help however the AI's second response may be just as wrong in a different more subtle way, so there is an argument in both directions. This could be part of a strategy to deal with this sort of problem but should not be their only answer
3. *Accepting this as the truth because the AI seemed to know what it was talking about.* ❌ - AI makes things up, even the slightest sliver of doubt should prompt a double check
4. *Asking a volunteer whether the AI is right.* ✔
5. *Asking another AI tool.* ❓ - This can be helpful but that tool might also struggle with the same issue so there's a case for asking a google or a human the moment AI feels unreliable.

## Question 3
*A CYF student balances fulltime work and family responsibilities with studying at CYF. They are struggling to keep up with the course. In particular, they are struggling to understand some fundamental concepts.*

*Which of the below are NOT appropriate responses?* 

1. *Asking volunteers for guidance at a session* ✔ - This is an appropriate space to ask for guidance. Volunteers can signpost them to staff, resources etc as well as helping solve the problem
2. *Giving an AI tool their assigned problems and ask it to solve them for them. They then copy and paste the output, making no edits.* ❌- this is plagiarism and does not promote their understanding 
3. *Asking an AI tool to explain the concepts they find difficult to understand.* ✔ - This is an appropriate use of AI
4. *Attempting to solve the problems, then giving up asking an AI tool to fix their code. Then copying and pasting the answer without understanding the changes it made.*❌ - this is plagiarism and avoids a key aspect of software dev - problem solving.
5. *Using Google to find articles on the topics they are struggling with.* ✔ - While "old fashioned" this is an appropriate way to solve the issue
6. *Copying another student’s answers.* ❌ - this is plagiarism
7. *Sending a direct message to a active volunteer* ❌- This is wrong because it means that this volunteer could be overwhelmed. This is why we encourage asking in channels from the very beginning. If they do have to ask something confidential it would be better to approach a relevant staff member
8. *Giving an AI tool their assigned problems and ask it to solve them for them. They then copy and paste the output, making small edits.*❌ - This is still plagiarism and they will not understand the code

illicitonion:
I think those answer keys are great - really clear, and really helpful.

I'd consider the phrasing of question 3 - your answer key here reads as if it's the answer to what _is_ appropriate but the question is what is _not_ appropriate.

Ara225:
I've revised that thanks. I think this pull request is good to go otherwise now?

illicitonion (approved):
LGTM, ship it! 🎉

Ara225:
ahhhhhhhhhhhhhhh I've just realized that this wasn't merged before I was lazy and deleted and recreated the fork. Luckily I was able to recover them and they are on my fork now, will just merge at the same time as the other AI step

PR #1527, opened 2025-07-31
Tweaks to SDC front page

illicitonion:
* Hide tracks menu, it's empty and not useful.
* Add a meaningful sub-title to the course page.
* Link to the Overview on the front page.


kfklein15 (approved):
Makes much more sense.

PR #1528, opened 2025-07-31
Improvements to CV guide

illicitonion:
* Re-order and re-title content to start with the "Sections of a CV" (which should come first), as well as a short introduction.
* Re-focus "previous careers" advice on positives rather than negatives.
* Add common CV mistakes, and fold the layout advice into it.

I think we have some room to keep improving this, but it focuses it a bit more, and gives us some natural FAQ-style extension points that are useful for linking to.


org-cyf-guides/content/employability/cvs/_index.md
```diff
@@ -4,3 +4,7 @@ title: CVs
 description: Learn how to tailor your CV to the right employer
 weight: 2
 ---
+
+The purpose of a CV is to show an employer that you would be a good person to work with.
+
+Your CV should tell a story about who you are and why you are a good person to work with.
```
kfklein15, line 10: Should we add a phrase of: "A recruiter will skim through your CV for around 10 seconds. So you must stand out with your CV."

kfklein15 (approved):
I suggested one small change if you think it's worth it. Otherwise looking amazing - much better and a good starting point.

PR #1554, opened 2025-08-14
Require Codewars username in Step 1

illicitonion:
Trainees are meant to sign up to Codewars in the Onbaording module, and start doing it in Structuring and Testing Data.

In my experience, trainees who are behind often have never signed up to Codewars.

Require that they sign up for step 1. We will probably also start requiring that they have done a certain number of kata as part of subsequent steps.

I have already edited the backlog ticket for Step 1 to indicate this.


kfklein15 (approved):
Makes sense.

PR #1556, opened 2025-08-24
Add next-class-day-plan shortcode and wire to ITP index (fixes #1119)

### LINKED ISSUE #1119: Feature: The next class is link

JohnnyRobs19:
Title: Add next-class-day-plan shortcode and wire to ITP index (fixes #1119)

## What does this change?
- Adds a shortcode (common-theme/layouts/shortcodes/next-class-day-plan.html) that computes the next likely class day-plan and renders a direct link.
- Wires the shortcode into org-cyf-itp/content/_index.md so the module index shows a single "Next class" link.
- Heuristic used: next Saturday -> match course start month -> compute week/sprint -> index into the "scheduled" menu. Falls back when no data available.

### Common Content?
- [ ] Block/s

### Common Theme?
- [x] Yes

Issue number: #1119

### Org Content?
Module | Sprint | Page Type | Block Type
---|---:|---|---
org-cyf-itp | N/A | Module index page (_index.md) | N/A

## Checklist
- [x] I have checked spelling/grammar
- [x] I have previewed my changes locally

Please review: @SallyMcGrath

## Before / After (artifacts to attach)
Before:
- Full-page screenshot (current UI)
image

- Editor screenshot or paste of org-cyf-itp/content/_index.md (pre-change)
image

- Small reproduction command snippet:

```bash
cd ~/code/curriculum/org-cyf-itp
sed -n '1,200p' content/_index.md > before__index.md
hugo server -D --bind=0.0.0.0 --port=1313
# open http://localhost:1313

After:
- Full-page screenshot showing the rendered "Next class" link
<img width="940" height="666" alt="image" src="https://github.com/user-attachments/assets/2844a32d-b6b7-4af7-b219-1a646c965c63" />

- The new shortcode file: common-theme/layouts/shortcodes/next-class-day-plan.html
<img width="807" height="165" alt="image" src="https://github.com/user-attachments/assets/567efb94-e81f-47fb-832f-07d839ba9132" />

- The updated file: org-cyf-itp/content/_index.md
<img width="940" height="410" alt="image" src="https://github.com/user-attachments/assets/e432ec3f-1df7-4608-96cd-8e2cbb2b97ba" />

- Preview commands:

```bash
cd ~/code/curriculum
git checkout -b updated-link-1119
# add the new files and commit
git add common-theme/layouts/shortcodes/next-class-day-plan.html [_index.md](http://_vscodecontentref_/0)
git commit -m "feat(next-class): add next-class-day-plan shortcode and wire to ITP index (#1119)"
hugo server -D --bind=0.0.0.0 --port=1313
# open http://localhost:1313


illicitonion:
Thanks for putting this together! It generally looks reasonable, but one thing I'm not sure about is the `scheduled` menu - what are you expecting to be added to it?

As far as I can tell, this pull request doesn't create a menu called scheduled - what's meant to be in it? It looks like it's meant to be each sprint?

We already have a menu called "syllabus" which contains each module, and a menu called "module" which contains each sprint within a module - could we use these existing menus, rather than having to introduce another one?

PR #1558, opened 2025-08-26
Allow horizontal scroll on cards

shieldo:
When viewing pages of the curriculum on a constrained viewport (causing some elements, typically inset code, to make horizontal scrolling available) and when using touch interactions, the gesture to move the page horizontally is interpreted as a previous page/ next page action, even if one is trying to scroll a horizontally scrollable element. This provides a short-circuit to this behaviour such that if the touch gesture starts within an element where horizontal scrolling is available, the page navigation action is avoided, allowing horizontal scroll. The reason for this is it is currently impossible to scroll these elements horizontally using touch because this page navigation action hijacks the browser.


illicitonion (approved):
Wonderful, thank you!

PR #1560, opened 2025-08-28
Fix multiple choice click capturing

### LINKED ISSUE #1410: Bug: multiple multiple choice quizzes not capturing clicks correctly
**Describe the bug**

Given https://sdc.codeyourfuture.io/tools/sprints/1/prep/#computer-quiz
When I click on any multiple choice quiz that is not the first quiz
Then it selects the first quiz anyway

Each quiz should be encapsulated and independent, even with multiple quizzes on one page.

**How could this be fixed?**

This was working when we deployed this module, so there's been some unknown change.  It might even just be some mad quirk on my machine so would appreciate someone reproducing.

If it's a real bug then I don't mind looking into the fix, which is almost certainly just better encapsulation. 

**Link to content**

**Additional context**

illicitonion:
## What does this change?

Different quizzes in the same page apparently share the same ID space for labels, even though they're in separate shadow roots.

Fortunately, the `<input>` tags are nested inside the `<label>` tags, so the `for` attribute isn't necessary, they implicitly bind to their contained tag.

So removing the `for` attribute here preserves the desired behaviour, while removing the confusion of which element is being targeted.

### Common Theme?

- [x] `common-theme/layouts/partials/multiple-choice.html`

Fixes #1410

### DIFF
--- a/common-theme/layouts/partials/multiple-choice.html
+++ b/common-theme/layouts/partials/multiple-choice.html
@@ -35,7 +35,6 @@
           <li class="c-quiz__answer">
             <label
               class="c-quiz__label"
-              for="answer-{{ $index }}"
               ><input
               class="c-quiz__input"
               type="radio"
--- a/common-theme/layouts/partials/multiple-choice.html
+++ b/common-theme/layouts/partials/multiple-choice.html
@@ -38,7 +38,6 @@
               ><input
               class="c-quiz__input"
               type="radio"
-              id="answer-{{ $index }}"
               name="answer"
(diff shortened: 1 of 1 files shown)


LonMcGregor (commented):
Tested locally and it works.

This will still end up with multiple of the same IDs on one page for each answer, which is technically not allowed. Leaving them in doesn't seem to pose any problems as far as I can tell, but removing them now as well might prevent other similar issues later.

LonMcGregor (approved):
Looks Good

PR #1565, opened 2025-09-02
Add extra PR to S&TD Step

illicitonion:
We split the coursework into two PRs, and we expect both of them in the step.


kfklein15 (approved):
LGTM

PR #1566, opened 2025-09-02
Add requirement to pass a Codility test

illicitonion:
Rather than having this be an entry criterion to the Piscine, we want to make it an exit criterion of ITP. This better reflects completing ITP meaning you can apply the skills learnt in ITP.


illicitonion:
Yeah, not yet decided - so definitely not the current cohort, but maybe the next one.

PR #1569, opened 2025-09-08
Add backlog to Welcome sprint

illicitonion:
We try to use the Welcome sprint as an example of how all sprints work.

But not having a backlog makes it unusual, and means it's a bad example.

Add a backlog with a couple of useful tasks to do before the welcome day.


kfklein15 (approved):
LGTM

PR #1573, opened 2025-09-13
Clarifications about Piscine interview next steps

illicitonion:


kfklein15 (approved):
I think from the conversation on Monday we kind of want to bring this forward either way. Is that a correct understanding? 

PS: We just have to communicate it to everyone :D

PR #1574, opened 2025-09-16
Add group work requirements

illicitonion:
This requires a screenshot on each of steps 2-4 showing that they worked together with others.

I'm not adding this to Step 1 because people are currently doing it (and some have already done it), but I'll add it to Step 1 next time around.


kfklein15 (approved):
LGTM

PR #1582, opened 2025-09-18
Add overview page to Piscine

illicitonion:
All courses should have an overview page we can link to, e.g. when introducing a course to new volunteers or hiring partners.


LonMcGregor (changes requested):
The page content looks good, however it seems that the only way to access it is to go to Prep > then click past the volunteer + success pages. Might it make sense for this to be the very first thing, either before or as the first section of prep?

If I compare to ITP, ITP has a button on the main page that takes you right to the overview

PR #1601, opened 2025-09-30
Update Piscine Sprint 1 Day Plan

illicitonion:
This fits more in line with what we actually do, and reflects that people are expected to come with a demo, and come knowing about the first sprint project.


LonMcGregor (approved):
This layout makes more sense now. Trainees get practice breaking things down, then do it "for real" with the project kickoff.

Adding a more explicit briefing at the beginning is a good idea also

PR #1661, opened 2025-11-19
Clean up Databases track

illicitonion:
This makes this a much easier to follow independent chunk of work. It removes a lot of leftover noise from when this was a module in the old course.

We will want to give a full review and improvement of this in time, but for now this makes it a useful stand-alone piece of work to follow.

Major changes:
* Removed all non-database content (e.g. job searching).
* Moved from Module-Databases to Module-SQL (a new repo) which has a very small number of issues, because there were previously so many unrelated things.
* Removed sprint 4 and squished things together a bit into 3 sprints.
* Removed all day plans. None of them contained any SQL-related content other than one workshop which the volunteers were meant to create a bunch of content for themselves.
* Make the track actually show up in the tracks list.
* Rename Databases -> SQL.

The relevant deploy preview is https://deploy-preview-1661--cyf-tracks.netlify.app/sql/

If we're happy with this, we should also archive https://github.com/CodeYourFuture/Module-Databases

### DIFF
--- a/org-cyf-tracks/content/sql/sprints/1/_index.md
+++ b/org-cyf-tracks/content/sql/sprints/1/_index.md
@@ -1,10 +1,10 @@
 +++
 title = 'Sprint 1'
-description = 'The plan for the sprint'
+description = 'Introduction to databases'
+theme = "Introduction to databases"
 layout = 'sprint'
 menu_level = ['module']
 weight = 2
-theme = "Introduction to databases"
 +++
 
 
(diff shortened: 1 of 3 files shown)


cifarquhar (approved):
Looks good! I think we should still have a broader discussion around leaving this as a standalone track vs incorporating some/all of it into another part of the course (probably ITP) but this is a much better starting point for whatever we do with it.

PR #1705, opened 2026-01-13
update backlog issue presentation

cifarquhar:
When the backlog pages are built learning objectives are stripped from the issues and nicely formatted with checkboxes, but the "Learning Objectives" heading is left behind. This removes the heading too.


LonMcGregor (approved):
This looks fine - It looks like it will leave the heading tag, but as long as the content itself is empty, it renders well.

PR #1730, opened 2026-02-01
Include links to original sources in success pages

illicitonion:
This:
* Makes it easier to see what things are self-study vs have expected evidence from backlog.
* Makes it easier to link back to where the topics were introduced.

The way we go about this is:
1. When fetching remote learning objectives, parse out to accumulate the list of LOs, rather than just grabbing the whole markdown block
2. Pass LOs through a `{{<objectives>}}` shortcode, rather than just pulling the shortcode in through the raw markdown.
3. Pass links to the original blocks into the rendering block.
4. Send both local and remote learning objectives through the same rendering block (`common-theme/layouts/partials/objectives/for-success.html`) which now includes links back.

### DIFF
--- a/common-theme/layouts/partials/objectives/for-success.html
+++ b/common-theme/layouts/partials/objectives/for-success.html
@@ -4,5 +4,5 @@
 {{ $title := $scope.title }}
 {{ $links := $scope.links }}
 
-{{ if $showTitle }}<h5>{{ $title }} {{range $links}}<a class="e-button" href="{{ .href }}">{{.text}}</a>{{end}}</h5>{{ end }}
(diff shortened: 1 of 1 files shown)


illicitonion:
> Am I the only one that ticked all the boxes to see if there was a done effect (not disappointed) ? Although it is not immediately obvious that they are interactive tick marks.

Yeah, it's more of an easter egg than a key feature (though we do generally demonstrate this when showing off the curriculum to people for the first time)

> I'm not sure I'm a huge fan of the Backlog etc hyperlinks. There are well-established UI conventions for hyperlinks and I'm not sure a green shadow which disappears when you hover over it is on the list. There a places for creativity, and others for convention.

This is our "button" style, rather than our "link" style. I've re-styled them as links rather than buttons, WDYT?

PR #1732, opened 2026-02-01
Add expand/collapse all button to backlogs

illicitonion:


OracPrime:
My comment on the other PR about buttons with green shadows not being a recognised convention for hyperlinks is underlined by this code, where the same UI convention is now *not* a hyperlink.  We should be consistent.

OracPrime:
Liking the new hyperlink style.  If it were me I'd drop the bottom-margin:10px on the list items - it just spreads things out too much.

PR #1773, opened 2026-03-06
Feature/trainee code review guide

### LINKED ISSUE #1651: Introduce trainees to reviewing code
At some point we should introduce some basic "How to review code" / "What to look out for" kind of content for trainees. Ideally their first experience of reviewing code wouldn't be on a job.

In particular, we should talk a bit about how to review code a GenAI tool has produced.

chinar-amrutkar:
Closes #1651.

### DIFF
--- a/org-cyf-guides/content/reviewing/code-review-basics/index.md
+++ b/org-cyf-guides/content/reviewing/code-review-basics/index.md
@@ -8,11 +8,6 @@ objectives=[
     "Apply constructive feedback in code reviews",
     "Review AI-generated code critically"
 ]
-[build]
-  render = 'never'
-  list = 'local'
-  publishResources = false
-
 +++
 
 # Code Review Basics


LonMcGregor (commented):
Good work!

Do you think it would be worth adding links somewhere near the end to the more detailed guidance we have, as an "optional extra" people could look at if they are interested in further reading?
* Our style guide, which serves as a good example of what trainees might encounter in the workplace: https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/
* the "what to look for" page which goes into more detail about specific problems that might be found during a review: https://curriculum.codeyourfuture.io/guides/reviewing/checklist/

Also, is this just a module at the moment, or is it going to be included in a specific place within the curriculum? I can't tell just from the PR. I ask because the link at the bottom points to the SDC slack - are we definitely intending this only to be used by SDC trainees? if ITP trainees could see this we might need a different link.

PR #1819, opened 2026-03-31
Add scroll indicator gradient to table of contents

### LINKED ISSUE #1799: Table of Contents is not obviously scrollable on mobile
The table of contents, when viewed on a mobile device, isn't clearly scrollable. For example, in SDC Tools, Sprint 3, there are 5 topics, but only 4 are visible. It isn't clear from the UI that the `c-toc` section has additional content or that it can be scrolled. To be clear, it can be **scrolled**, but it isn't obvious.

Image

This is a usability / discoverability issue for users.

**How could this be fixed?**
Some possible solutions:
- Partial content cut off: show part of the next line of text
- Faded content at the bottom of the section
- An always visible scrollbar
- A scroll indicator icon

**Link to content**
 SDC Tools, Sprint 3 is an example, but it could happen elsewhere:
https://sdc.codeyourfuture.io/tools/sprints/3/prep/

chinar-amrutkar:
On mobile, the TOC container clips content at max-height without visually indicating there is more below. Adds a CSS gradient fade at the bottom to signal scrollable content.

Fixes #1799

### DIFF
--- a/common-theme/assets/styles/04-components/toc.scss
+++ b/common-theme/assets/styles/04-components/toc.scss
@@ -9,18 +9,17 @@
     var(--theme-spacing--4);
   overflow-y: auto;
   counter-reset: item;
-  position: relative;
-
   // Gradient fade at bottom to indicate scrollable content (closes #1799)
   &::after {
     content: "";
-    position: absolute;
+    position: sticky;
     bottom: 0;
-    left: 0;
-    right: 0;
+    display: block;
     height: 2em;
+    margin-top: -2em;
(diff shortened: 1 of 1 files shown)


illicitonion (commented):
This looks good to me, but it looks like the gradient stays in place as you scroll:

Image

Is there an easy way to make the gradient fixed to the bottom? (And ideally fade out if there's no overflow, but that feels probably more fiddly!)

illicitonion (approved):
Thanks!

PR #1827, opened 2026-04-13
Remind readers to check they've met learning objectives

illicitonion:
This shows up on every day-plan block which has learning objectives (but not tasks).

Currently only supported for local and readme blocks, but I think that's all of the relevant ones.

Doesn't show up on non-day-plan pages (e.g. prep).


Ara225:
Looks good to me never hurts to remind the trainees to reflect

PR #1830, opened 2026-04-14
Swap onboarding tasks

illicitonion:
Form Controls is easier and less review-interesting than wireframe, so we're going to swap their weeks. Accordingly, move around the prep a little.

I will re-label the issues just before merging.


LonMcGregor (approved):
This sounds like a good change. Looks fine to me

PR #1832, opened 2026-04-20
Tweaks to ITD S&TD Sprint 1

illicitonion:
Be more specific about how the screen safari should work in-person.

Also, use the same title for the question workshop in the menu and the heading.


Poonam-raj (commented):
Lovely

Poonam-raj (approved):

PR #1840, opened 2026-04-23
Merge multiple backlogs together

illicitonion:
We support pulling in issues from multiple backlogs - currently they end up being shown separately (with separate "Setup"/"Mandatory"/... sections).

Instead, merge them into one list of issues before rendering.


cifarquhar (approved):
Looks good!

Issue #1848, opened 2026-04-30
Updates to ITP Onboarding


cifarquhar:
**Which module(s) and week(s) does this change affect?**
Module(s): Onboarding
Week(s): All

**What is the work that needs to be done?**

We are reorganising the first module of ITP to include JavaScript content. The majority of content will be pulled forward from Structuring & Testing Data and updated as necessary but some new material will be needed. Details of the proposed new structure are on a [Miro board](https://miro.com/app/board/uXjVGmM1MYA=/?share_link_id=890596212970), contact @cifarquhar for access if needed. 

Sub-issues will be created with details of specific tasks and linked from here.

**Why is this work important to do?**

Onboarding is currently very light on content and doesn't give an accurate representation of what future weeks will look like. By adding more to the prep work trainees have a better idea of what they're signing up for before enrolling.

This also gives us an opportunity to spend time consolidating key topics such as Git usage, and opens up space in later modules to add more content there.

**How to contribute**

Contributions are welcome, but before picking something up please bear in mind:

- Work is being done on the `onboarding-updates` branch
- These changes are more involved than just adding a single piece of content or updating some code. They will fit together to form a coherent narrative through the module so communicating with other contributors will be critical. The [deploy preview](https://deploy-preview-1849--cyf-curriculum.netlify.app/) will show the current state of the module, but please ask questions too!
- Most files which have been moved from elsewhere have a red `CAUTION` box at the top highlighting potential issues
- New files have suggested key points in a bullet-point list. Feel free to adjust these as necessary, or to split/combine files if it makes more sense in context.

PR #1849, opened 2026-04-30
ITP Onboarding Updates

cifarquhar:
This PR will reorganise the Onboarding module of ITP to bring more content into the first three weeks of the course. It will also fill gaps created in later modules by moving things forward.

[This issue](https://github.com/CodeYourFuture/curriculum/issues/1848) links to details of planned changes.

**Progress**:

- [X] High-level restructure of Onboarding prep
- [x] Write notes for new prep content
- [x] Redraft old content to align with new structure
- [ ] Move backlog items $^1$
- [ ] Reorganise workshops
- [ ] Fill gaps in later modules $^2$
- [ ] Consistency checks

1 - Backlog changes will require a change in how we store tasks - working on it!
2 - Currently only Structuring & Testing Data, but this may change


common-content/en/module/js1/comparison/index.md
```diff
@@ -3,143 +3,95 @@ title = 'Comparing current and target output'
...
-```
+We have learned how to store values in a program and how to access them again when we need them. At some point we will need to **compare** two values. Often our program will need to change its behaviour depending on that comparison. 
 
-Suppose we now call the function like this:
+For example, think about what happens when you switch on your laptop. Before you do anything else you are asked to enter a password. Whatever you type is compared against a value stored on the computer and if the two match you can carry on using your computer. If they are different you won't be able to access files or use any programs.
 
-```js
-formatAs12HourClock("05:30");
-```
+We use comparison operators to compare two expressions. We use the equality operator `==` to check if two values are the same.
```
illicitonion: I'm a bit worried that we're introducing `==` as the ideal here, and that we're mis-describing it below in the first example ("using the strict equality comparison expression" with `==`).

I would be tempted to invert this - introduce `===`, give the examples using `===`, then have one at the end that is `4 === "4"`, show that it's false, and comment on how this kind of comparison is rarely what we want but FYI `==` exists if you really want to do it but please don't

common-content/en/module/js1/define-computer/index.md
```diff
@@ -13,4 +13,5 @@ time = 5
 
 Modern computers are **complicated**: it would be too difficult and time-consuming to list all the components that make up a modern computer. So to build our mental model, we will use this simple definition of a computer:
 
+>[!Definition]
```
illicitonion: Does this work?

common-content/en/module/js1/errors/index.md
```diff
@@ -28,21 +26,21 @@ On line 1, we have a variable declaration, but the string has a missing `"` We'r
...
-```node
+```console
 const firstName = "Francesco;
                   ^^^^^^^^^^^
 
 Uncaught SyntaxError: Invalid or unexpected token
 ```
 
-We get a **SyntaxError** message. This error message is telling us that we've broken the rules of the language.
+We get a **SyntaxError** message. This error message is telling us that we've broken one of the rules of the language. In this case the interpreter didn't expect to see the semicolon - it needs us to add the missing `"` before the expression makes sense.
```
illicitonion: I think technically the unexpected token here is the newline rather than the semicolon?

```suggestion
We get a **SyntaxError** message. This error message is telling us that we've broken one of the rules of the language. In this case the interpreter didn't expect to see the end of a line - it needs us to add the missing `"` before the expression makes sense.
```

common-content/en/module/js1/functions/index.md
```diff
@@ -3,58 +3,60 @@ title = 'Functions'
...
   list = 'local'
   publishResources = false
 
 +++
 
-Consider the number `10.3`.
+When we are writing programs we often find ourselves needing to do the same thing over and over again. Think back to our password checker from the previous sprint: there are lots of places where you need to enter a password!
 
-> 🤔 "What is the nearest whole number to `10.3`?"
+We could re-write the code to check a password every time we needed to use it but that wouldn't be very efficient. It would take a long time to write and there's a chance we could make a mistake and introduce a bug. It would be much easier if we could write the code once and reuse it anywhere it was needed.
```
illicitonion:
```suggestion
We could re-write the code to check a password every time we needed to use it but that wouldn't be a good use of our time. It would take a long time to write and there's a chance we could make a mistake and introduce a bug. It would be much easier if we could write the code once and reuse it anywhere it was needed.
```

common-content/en/module/js1/interpreting-errors/index.md
```diff
@@ -12,86 +12,105 @@ time = 20
...
-
-Node.js v22.4.1
+% node passwordChecker.js
+
+/Users/yourname/cyf/passwordChecker.js:11
+  response = "Incorrect password!";
+           ^
+
+TypeError: Assignment to constant variable.
+    at Object.<anonymous> (/Users/colinfarquhar/clients/cyf/content_testing/passwordChecker.js:11:12)
```
illicitonion:
```suggestion
    at Object.<anonymous> (/Users/yourname/cyf/passwordChecker.js:11:12)
```

common-content/en/module/onboarding/goals/index.md
illicitonion, line 1: This file talks as if it's in the old Onboarding module - probably needs some rework to reference the JS Fundamentals modules and timings instead?

org-cyf/content/itp/onboarding/prep/index.md
```diff
@@ -0,0 +1,66 @@
...
+name="Development Process"
+src="module/induction/development-process"
+[[blocks]]
+name="Version Control"
+src="module/induction/version-control"
+[[blocks]]
+name="Sharing History"
+src="module/induction/sharing-history"
+[[blocks]]
+name="Configuring Git with VSCode"
```
illicitonion, line 58: This block has "how to install git" instructions but comes after "Check Git Installation" - let's remove the "how to install git" from this one?

common-content/en/module/onboarding/git/remote-repositories/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+{{<note type="info" title="Why not initialize with README?">}}
+If you initialize with files on GitHub, your local and remote repositories will be different, which causes conflicts. Since we already have commits locally, we'll connect them directly.
+{{</note>}}
+
+### Step 2: Add the Remote Connection
+
+GitHub will show you a page with instructions. You need to tell your local Git about this remote repository.
+
+First copy the url to your clipboard.
```
illicitonion: If we're actively having people use SSH, should we call out here to make sure they click "SSH" and that the URL starts with `git@` not `https://`?

Or do we want them to use `https://` and configure auth that way? In which case, let's update these screenshots to do that?

org-cyf/content/itp/javascript-fundamentals/sprints/1/prep/index.md
```diff
@@ -64,15 +25,29 @@ src="module/induction/viewing-files"
...
+name="Fork your Planner"
+src="https://www.youtube.com/watch?v=cnx0RuAu2tc"
+time=5
+[[blocks]]
+name="Make your Planning Board"
+src="https://www.youtube.com/watch?v=Hbtfil-G0h0"
 time=15
 [[blocks]]
+name="Set up Planner"
+src="https://github.com/CodeYourFuture/Coursework-Planner/tree/main"
```
illicitonion: Do we want to remove the clone instructions from here, or leave them for now?

org-cyf/content/itp/javascript-fundamentals/sprints/2/prep/index.md
```diff
@@ -0,0 +1,60 @@
++++
+title = 'Prep'
+layout = 'prep'
+menu_level = ['sprint']
+weight = 1
+[[blocks]]
+name="Spaced Repetition"
+src="blocks/spaced-repetition"
```
illicitonion: I think we should probably just remove this. I would love for us to add a meaningful spaced repetition element to the course, but it isn't there right now, and without context this feels quite out of nowhere?

common-content/en/module/js1/scripts/index.md
```diff
@@ -1,47 +1,88 @@
...
   list = 'local'
   publishResources = false
 
 +++
 
-So far we’ve seen how expressions can be evaluated using the Node REPL. The Node REPL is useful for evaluating expressions quickly.
+It's time to write our first lines of JavaScript! 
 
-But usually, our programs have many instructions, and we want to keep and re-run them instead of typing them out each time. So we save our instructions in files. Node can also execute instructions written in a file.
+There are some tools available which will enable us to write code and instantly see the results. These are called **REPL**s - **R**ead, **E**valuate, **P**rint and  **L**oop. These are great for quickly checking something but not very practical for production uses.
```
illicitonion: Let's maybe remove the REPL definition here? It feels weird to introduce them as the first way to run code and then not use it?

common-content/en/module/js1/conditionality/index.md
```diff
@@ -40,54 +38,96 @@ flowchart LR
...
+We're going to write some code to implement our password checker example. We'll use `console.log` to print messages for the user which will change depending on the password they enter. Let's start by creating a file called `passwordChecker.js` to work in and initialising some variables.
 
-```js
-function checkDivisibility(a, b) {
-  if (a % b === 0) {
-    return `${a} is divisible by ${b}`;
-  }
+```js {title="passwordChecker.js"}
+const password = "secretword123";
+const userInput = "secretword123";
```
illicitonion, line 51: I wonder if we want to explicitly call out in this block that in real life we'd receive the user input from a user rather than write it as a literal ourselves, but that even if we do that the rest of the code will be the same, and maybe even introduce the idea of hard-coding (in a very hand-wavy way)?

org-cyf/content/itp/javascript-fundamentals/sprints/3/prep/index.md
```diff
@@ -0,0 +1,43 @@
++++
+title = 'Prep'
+layout = 'prep'
+menu_level = ['sprint']
+weight = 1
+[[blocks]]
+name="Enroll as a Trainee"
+src="https://github.com/CodeYourFuture/Module-Onboarding/issues/72"
```
illicitonion: We'll need to update this, but I don't know what to yet

common-content/en/module/onboarding/git/undoing-a-commit/index.md
illicitonion, line 1: Is it worth adding a caveat here that you can't do this if you've already pushed, and that if you've done that, you'll need to do something else?

Also is it worth introducing "reverting" in contrast to "undoing"?

common-content/en/module/onboarding/git/undoing-a-commit/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Undoing a commit
+
+Let's revisit the [educational blog](https://github.com/CodeYourFuture/education-blog) we looked at in Sprint 1. Open the folder in VSCode and take a moment to refamiliarise yourself with the files.
+
+Let's imagine a world where we're putting this page online using a platform like GitHub Pages. usually we need to include some sort of configuration information when we deploy an application, typically things like GitHub urls to load content from or passwords for third-party services. We're going to add a fake file which will store some credentials for our imaginary deployment.
+
+Create a file called `passwords.json` at the directory's root and add the following object to it:
+
+```json {title="passwords.json"}
```
illicitonion: Let's update this to just use a `.env` file rather than a JSON file - we haven't introduced JSON yet in the course, so adding a new file format, and then in the next block saying "Oh but we'll do it another way" feels needlessly complex?

common-content/en/module/js1/interpreting-errors/index.md
cifarquhar: Resolved by #1986

org-cyf/content/itp/javascript-fundamentals/sprints/1/prep/index.md
cifarquhar: Done - now covered in a broader planner setup session

common-content/en/module/onboarding/git/remote-repositories/index.md
cifarquhar: Good spot, thank you! 

I'm honestly not sure on this, SSH feels like it has the potential to be a significant blocker. The next section suggests taking the time to set it up but doesn't say it's mandatory. I could bring that forward and add SSH setup to an earlier section. Thoughts?

common-content/en/module/onboarding/git/undoing-a-commit/index.md
cifarquhar: Good point on pushing, I'll add a note on that.

In terms of calling out "reverting" specifically (or indeed "resetting" too), the issue here is that VSCode is very limited in terms of how much you can undo with Git without adding various extensions. When I add the CLI content I'll go deeper on the difference and show how to do each.

common-content/en/module/onboarding/git/remote-repositories/index.md
illicitonion: I _think_ VS Code's Git generally supports "log in one time and it'll auth over HTTPS", so I'd lean towards just doing that?

If not, let's actively walk them through setting up an SSH key. Which I don't hate either :)

common-content/en/module/onboarding/goals/index.md
cifarquhar: Made some changes in a way that I _think_ makes more sense.

org-cyf/content/itp/onboarding/prep/index.md
cifarquhar: Good spot, thank you! Done.

org-cyf/content/itp/javascript-fundamentals/sprints/2/prep/index.md
cifarquhar: Done

common-content/en/module/js1/scripts/index.md
cifarquhar: Done

common-content/en/module/js1/conditionality/index.md
cifarquhar: Done - added a definition of hard-coding to say we're only doing it for demo purposes

illicitonion (approved):

Issue #1706, opened 2026-01-13
Expand description of array iteration


Poonam-raj:
**Which module(s) and week(s) does this change affect?**
Module: ITP - Data Groups
Week: 1

**What is the work that needs to be done?**

Prep material on [iteration](https://programming.codeyourfuture.io/data-groups/sprints/1/prep/#iteration) lacks fundamental ideas of iteration. We show them `for...of`, but there is space here to show them or explain the [fundamental for loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for) more fully. I don't believe it is shown before this point. 

Add an introduction to the idea of iteration more generally, including:

- That it is a general concept across coding languages.
- That our most fundamental loop is a [for loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for).
- that other loops exist and docs are great for that. 

**Why is this work important to do?**

Fundamentals are the foundation on which good problem solving can occur, as we can always go back to those fundamentals when we want to truly shape an algorithm to what we want. It would be useful for us to reference those fundamental building blocks such as a `for` loop at this early point in learning about data structures.

PR #1869, opened 2026-05-01
expand iteration explanations for array for and for of loops

Poonam-raj:
- Expands general iteration explanation
- dives deeper into `for` and `for...of` loop


Poonam-raj:
Addressing issue: https://github.com/CodeYourFuture/curriculum/issues/1706

cifarquhar (approved):
This is great! I like the having the step-by-step before and after the example, it gives some good context for the explanation

PR #1874, opened 2026-05-06
Clarify which email

illicitonion:
People often us the wrong one. This may or may not help, as often people don't remember which they used, but at least it's something!


NJ-G (approved):
A clearly described nudge; displays well on desktop and mobile.

PR #1841, opened 2026-04-23
Expose course data over a (static) API

illicitonion:
This allows other systems to accurately model our courses.

There are currently no guarantees around stability of this API.

Depends on https://github.com/CodeYourFuture/curriculum/pull/1840

You can see the output of this at https://deploy-preview-1841--cyf-curriculum.netlify.app/course_data/index.json

Issue #1822, opened 2026-04-06
Merge workshop content into currlicum


Ara225:
As discussed on the doc at https://docs.google.com/document/d/1lNlPgClo1gkW_j6kW4pf6HUivCiRkqa6XrR5tVcHF7A/edit?tab=t.0 we are planning to move all the workshops into the curriculum repo. 

I have created a temporary repo  to place code examples etc into - will need to move that to the orgif this way of organizing looks good. Not all of the workshops in the workshops repo are referenced in the curriculum  - in fact most aren't - I propose either archiving it or scoping the permissions down so noone can commit to it (to retain discoverability through search if that is useful)

Had to make some small changes to the workshop wording to send people to the right place

Questions/notes:
- The Get Forms workshop requires that branch preview deploy are turned on for forks of whatever repo it's HTML file is hosted in - seems to be because of the need to swap projects. This is a proper annoying one - risky if any secrets are lying around so better not to enable on the curriculum repo ... maybe something could be done with github pages or somat? Needs a little thought
- Didn't remove references to workshops in examples - need to review significance
- Need to move repo into org if that way of organizing suits
- express-setup has code in own repo, a fork of another project - it is only referenced in the servers track
- Should the time attrs move inside the workshop markdown file?
- Where can we document what we are doing with the workshop code?
- The .questions folder - not a workshop but a complete mystery so far
- node-challenge-london-mini-guide is weird It links to a web app that I can't see in our netlify (may ofc be a permissions thing but I can see loads) THere is a repo called the same name but it has no code. However it is just a demo of what they should do and isn't really 
necessary so I guess it is fine. 
- With the node-challenge-london-mini-guide, sample code is included in the folder, workshop doesn't refer to it at all

Progress:
arrays ✅
asking-questions ✅
crud ✅

**debugging**
- Code including the content ✅
- Static content - N/A
- markdown file ✅
- Readme.md renamed to index.md ✅
- All links in content changed✅
- link to content✅

devtools
- markdown file ✅
- Static content✅
- Code - N/A
- Readme.md renamed to index.md ✅
- All links in content changed ✅
- link to content ✅

dom-merge-conflict
- markdown file ✅
- Static content - N/A
- Code - ✅
- Readme.md renamed to index.md ✅
- All links in content changed ✅
- link to content ✅

express-setup
- markdown file ✅
- Static content - N/A
- Code - N/A
- Readme.md renamed to index.md ✅
- All links in content changed ✅
- link to content ✅

[...]

PR #1878, opened 2026-05-08
Json backlog testing

cifarquhar:
**DRAFT - Testing an alternative way of storing backlog issues**

- Adds `org-cyf/data/backlog.json` with title, category and url of current backlog issues
- Updates `issues-list-as-blocks.html` to load the list of issues from this file instead of using the GitHub API
- Updates `backlog.html`, `backlog-issue.html` and `block.html` to reflect the different data structure
- Updates `backlog/index.md` for each ITP/SDC module to update labels
- Removes the [TV Show Sprint](https://curriculum.codeyourfuture.io/itp/data-flows/project/) as these issues are now included across Data Flows
- Updates `success.md` to display backlog titles correctly. Note: **this means workshop titles do not display correctly**. Leaving this for now while work is ongoing on #1822, I'll revisit once we have all the workshops in the same format.

Currently this works with a copy of the JSON generated by #1841. If a module wasn't included in that then it won't work here.


common-theme/layouts/_default/backlog.html
```diff
@@ -34,19 +34,25 @@
     */}}
     {{ range $repos }}
       {{ $issueBlocks := partial "block/issues-list-as-blocks.html" (dict "backlog" . "backlog_filter" $backlog_filter "path" $currentPath) }}
+      {{ if eq 0 ($issueBlocks | len)}}
```
illicitonion, line 37: This is great for testing partial progress, but let's turn it into an error before merging

common-theme/layouts/partials/block/issues-list-as-blocks.html
```diff
@@ -1,69 +1,31 @@
...
-{{ $variables := dict
-  "owner" site.Params.owner
-  "repo" $repo
-}}
-{{ $currentPath := .path }}
-{{ if .backlog_filter }}
-  {{ $variables = merge $variables (dict "labels" .backlog_filter) }}
+
+{{ $courseRegex := `/(.*?)/` }}
+{{ $courseName := (trim (index (findRE $courseRegex .path 1) 0) "/") }}
```
illicitonion, line 12: I think  `.Page.Section` is the name of the course without needing to do path regex

illicitonion (commented):
LGTM!

Presumably at some point we'll inline the issue text (so we don't have issue URLs any more), and do some auto-magic around generating e.g. the "how to submit" text rather than having it inline in each issue? (But before doing that, we either need to update the issue cloner to create issues from our source of truth rather than issues, or need to stop using coursework planners like we are)

PR #1887, opened 2026-05-19
correct wording for GitHub link text

Liam310:
The text content of the link in question is actually "Browse files" rather than "Browse the repository at this point in the history" (the latter is in fact the text content of the tooltip that appears on hover). This PR amends that.

### DIFF
--- a/common-content/en/module/induction/checkout-a-commit/index.md
+++ b/common-content/en/module/induction/checkout-a-commit/index.md
@@ -20,8 +20,8 @@ We can use the Github interface to check out the files and folders at a previous
 
 Go back to this page https://github.com/CodeYourFuture/education-blog/commits/main
 
-Locate the the commit with hash `4e78b32` and then look for the icon that that says "Browse files".
-Explore the code at this point in the history. What differences do you notice?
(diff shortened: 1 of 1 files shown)


illicitonion:
Interesting - where do you see this text? When the screen is narrow I see:

Screenshot 2026-05-19 at 18 05 57

And when wide, I see no text except this tooltip:

Screenshot 2026-05-19 at 18 03 02

Liam310:
> Interesting - where do you see this text? 

Ah interesting, it would seem we are looking in slightly different places! I think I interpreted the instructions differently (possibly incorrectly, now I read them again). What I did was actually click through the commit to get to this page, where you can see the link with text content "Browse files" and tooltip text "Browse the repository at this point in the history".

image

I still feel that the instructions could be clarified a _little_ but I'm happy to revert my particular changes in favour of something slightly different.

illicitonion:
Makes sense! Give me a wave when you've made your tweaks!

Liam310:
@illicitonion I've made the changes, feel free to review at your leisure 🙂

illicitonion (approved):

PR #1892, opened 2026-05-25
The full Alfa java course 

Ara225:
This builds and I can't see anything obviously wrong apart from the fact that backlog issues aren't showing up on my local ( I added the relevant label) and this this warning that doesn't make a lot of sense:

WARN  "/home/anna/git_linux/curriculm_NOT_A_FORK/org-cyf/content/tracks/java/success/index.md:1:1": duplicate menu entry with identifier "End of Module Review" in menu "success criteria"

 And there is the TODO they left  in topic 4:  Alfa Streams and Lambdas exercises - TODO when we move this to git: need the superclass and the imports, as well as the source for NameFinder, ShapeCounter, Remainders..


illicitonion:
I've skimmed this, not gone word-by-word, but it looks great! I pushed a few small commits, I hope you don't mind, feel free to review them and complain!

A range of thoughts:
* We probably won't follow the day-plans precisely, that's ok, we'll see how they go and evolve them
* We've ended up with extra sprints... I see how this happened - in the doc we had a sprint 2b... I will ask the Alfa folks about this :D
* Sprint 2b/3's backlog seems missing? I guess this goes back to the sprint 2b vs 3 question?
* Sprint 8 is missing a backlog? I'm assuming that's missing from the source material from Alfa too?

Ara225:
Hi,
Thanks for that, good catches

Yeah the day plans and the timing for the prep items probably aren't that great but the source is a bit thin on details in that bit - probably best worked out in practice anyway 

There doesn’t seem to be a sprint 2b in the word doc for me, just 1-8,  could we have diffrent versions? 

On sprint 3 they've just written a sentence saying to submit excercises as the coursework, so I created a stub ticket saying that, wasn't 100% sure how to handle the line bluring and that seemed the least fuss.

Sprint 8 stops where excercises should be for the last topic so there's no backlog present

illicitonion:
All makes sense - I was indeed looking at an older draft

For the sprint 3 exercises - can you file a ticket like https://github.com/CodeYourFuture/Module-Tools/issues/178 which can act as a placeholder?

For sprint 8, let's worry about this after we merge!

I'll pull in the Alfa folks to take a closer look, then let's get this thing going!

Ara225:
@illicitonion Sorry for the delay in getting back to you. Sounds good, edited the stub issue I created to be in that format https://github.com/CodeYourFuture/Module-Java/issues/5#issue-4519730200

illicitonion (approved):
Let's merge this and get it deployed - we can make any edits we need in the future :) Thanks again so much for putting this together!

PR #1894, opened 2026-05-28
Update course descriptions

cifarquhar:
Finally fixing a couple of things that annoy me every time I look at the home page, and which I have to warn people about in every training session I run:
- Fixed the ITP course description so it no longer references things we don't teach
- Removed the duplicate "Post-ITP Tracks" block and renamed the other one "Specialist Tracks"
- Updated the Checkpoint emoji, because now the fish makes even less sense


illicitonion (commented):
Generally LGTM, but not sure on removing Post-ITP tracks - the "before checkpoint" vs "after checkpoint" separation does feel useful... WDYT?

cifarquhar:
Agree that the separation could be useful, but only if we make a distinction between which tracks are suitable for post-ITP and which are post-SDC. Currently it's the same list in two places, which is more likely to cause confusion. 

If we do keep them on a single list I think after checkpoint is better. They're courses designed to support people in finding specific roles, so listing them under "Intro Courses" doesn't feel right to me.

illicitonion (approved):
Fair enough! If only we could just express everything in life as a directed graph.

PR #1906, opened 2026-06-05
restructure instructions for working locally

### LINKED ISSUE #1888: Reorder the instructions for cloning a repo
The instructions for [how to work locally on a GitHub repo](https://curriculum.codeyourfuture.io/itp/onboarding/sprints/1/prep/#working-locally) seem to be presented out of order. For example, there is a diagram that breaks down the process of forking and cloning before cloning has been defined. Then there is a caveat to the task of cloning for users of a library computer before the task itself has been outlined.

The content itself is fine but it's just a bit jumbled.

Liam310:
- Reorders the page's content to flow more naturally, e.g. the diagram describing the process is placed at the end as a useful reference, not as an initial explainer
- Reorganises the work to follow a more typical structure, i.e. concept intro followed by `<note type="exercise">` block (rather than a more generic headed section)
- Adds a top and bottom `margin` to the `<youtube>` shortcode, otherwise there was no gap between the video and the exercise which didn't look good. I'm hoping that margin collapse will mean this makes minimal difference to videos embedded elsewhere on the site but I didn't check every single one so please correct me if I'm wrong!

Closes #1888

### DIFF
--- a/common-content/en/module/induction/working-locally/index.md
+++ b/common-content/en/module/induction/working-locally/index.md
@@ -12,53 +12,53 @@ time ="20"
 
 +++
 
-Here is a diagram representing how the repositories interact after forking and cloning:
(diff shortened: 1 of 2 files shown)


cifarquhar (approved):
Looks good! The video margin seems ok elsewhere, I picked a couple at random and couldn't see any issues.

PR #1912, opened 2026-06-11
Fix blockquote css

cifarquhar:
Updates the CSS for blockquotes to make them appear more consistent with the rest of the content. I have no strong opinions on the colour/size/padding so alternatives welcome 😃 

Also removes the dot pattern background from code blocks as it never lines up properly and just looks a bit weird.


cifarquhar:
@illicitonion This might fix what we were talking about the other day, example screenshot attached
Screenshot 2026-06-11 at 17 04 42

Image taken from one of the updates I was working on, the notes are on the `onboarding-updates` branch

illicitonion (approved):
LGTM!

We also use `dots` in `common-theme/assets/styles/03-elements/misc-phrasing.scss` in `.highlight` which is used in exactly one place (runkit codefences) - I think we're ok to just remove `dots` from `.highlight` too, but we probably should do this if we're removing the implementation of `dots` :)

cifarquhar:
Done, thanks!

PR #1918, opened 2026-06-22
feat(backlog): usability improvements, navigation fixes, and github edit buttons

### LINKED ISSUE #552: Feature:  open an issue, edit on github buttons

### LINKED ISSUE #1061: ---s on backlog pages get rendered with an incorrect left margin
The Code Review task on https://programming.codeyourfuture.io/data-groups/sprints/1/backlog/ renders:

Screenshot 2024-10-10 at 16 20 57

It feels like those `---`s should be within the rendered box?

### LINKED ISSUE #1072: Pressing left and right arrows in a text field in the prep view does navigation
On https://programming.codeyourfuture.io/data-groups/sprints/3/prep/#character-limit type something in the textarea, then press left or right to move the cursor so you can edit the content.

See that instead of interacting with the textarea, you end up navigating to a different prep section.

Our nav shouldn't interfere with browser-native controls like this.

JHONR2:
Closes #1061, Closes #1072, Closes #552

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ yarn.lock
 # Local Netlify folder
 .netlify
 .env
+.atl/
--- a/common-theme/assets/scripts/solo-view.js
+++ b/common-theme/assets/scripts/solo-view.js
@@ -141,6 +141,9 @@ class SoloView extends HTMLElement {
   }
 
   handleKeydown = (event) => {
+    if (event.target.closest('input, textarea, select, [contenteditable]')) {
+      return;
+    }
     if (event.key === "ArrowLeft") {
       this.navigateBack(event);
     }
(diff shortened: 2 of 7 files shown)


illicitonion:
Hi! Thanks for putting this together!

Can I ask how you tested this? I still reproduce #1072 with your change, and I don't currently see any pages suffering from #1061 before or after this change :)

The fix for #552 looks good - please can you re-style so that there isn't a big triangle before the cog, and it's instead just a cog you can click to expand?

PR #1924, opened 2026-06-30
docs: point devs to the running-locally guide

abdishakoor-dev:
The missing-token build error tells you to "Check the README", but the README had no local setup instructions. This adds a pointer and fixes a version mismatch.

- **README:** add a "Running locally" section linking to the running-locally guide
- **github-auth.html:** change the missing-token error to point at the running-locally guide instead of the README
- **running-locally guide:** correct the required Hugo version (`0.136.5` -> `0.136.3`) to match `netlify.toml`


cifarquhar (approved):
All looks good, thank you!

PR #1933, opened 2026-07-01
Disable Register button while submission is in-flight

Joshuafolorunsho:
Prevents double-submits by disabling the button, swapping its label to "Saving…", and switching the cursor to not-allowed until the request resolves. On error the button and its label are restored so the user can retry.

https://github.com/user-attachments/assets/3a1971f7-11ba-4d1f-9f5c-a071cca0b9e8


illicitonion (approved):
Perfect, thank you so much!

PR #1934, opened 2026-07-03
SDC demos can be 5 minutes not just 2

illicitonion:
We may want to extend the time in the day-plan we give to demos too, but for now, just give a clearer indication of the time expectation.


Poonam-raj (approved):
Makes sense

Issue #1441, opened 2025-05-09
Add a "Breaking down a project into tickets (and unblock each other)" workshop to ITP


illicitonion:
In the Checkpoint we currently give groups projects to do without breaking them down into tickets.

Checkpoint is only meant to test existing skills, not teach new ones. But we haven't taught this skill in ITP. So for the first few Checkpoints we've had to give [a workshop on how to do this](https://curriculum.codeyourfuture.io/checkpoint/sprints/1/day-plan/#practice-breaking-down-a-requirement). Ideally we would like to stop teaching this in Checkpoint and start teaching it in ITP.

We have a similar workshop in the SDC (https://sdc.codeyourfuture.io/decomposition/sprints/1/day-plan/#breaking-down-features-workshop) but it's focused on a specific project and is specifically talking about backends and frontends, which is a bit too much for ITP.

We probably want to use "if we can agree on the keys in an object" or "if we can agree how to serialise to `localStorage` and then how to deserialise" or something as our example.

Some of the natural places this could fit (we could do more than one):
* In Data-Groups we assign [the tic-tac-toe project](https://github.com/CodeYourFuture/Project-terminal-tic-tac-toe) which is done in groups. We have _already_ broken this down for people so that they can work in separate functions. After completing this, we could have a reflection session: "Here's how we broke it down so you could work in parallel".
* In Data-Flows we have [a Teamwork Project](https://curriculum.codeyourfuture.io/itp/data-flows/sprints/3/day-plan/#teamwork-project) (which is just talking about doing a project, not actually doing it) where we introduce the idea of dependencies between tasks. I'm imagining we could add an "identifying and breaking dependencies" workshop this morning before the teamwork project which could influence the teamwork project, or augment the teamwork project to cover this.

illicitonion:
We discussed doing this with the Tic-Tac-Toe project which has already been broken down for people.

PR #1939, opened 2026-07-08
Add 'Breaking down a project to work in parallel' workshop (ITP, #1441)

abdishakoor-dev:
First pass for #1441 

An ITP workshop on breaking a project into tickets for parallel work (so it's taught in ITP, not Checkpoint). Reflects on how Tic-Tac-Toe was broken down, teaches the steps, then groups practise on a fresh `localStorage` requirement. Added to the Data-Groups **Sprint 2** day-plan (~50 min, from Study Group).

Preview: `/itp/data-groups/sprints/2/day-plan/`

Questions/conerns:
- Is Sprint 2 + ~50 min the right placement/budget? Morning sessions already have some workshops/handson, and this one feels heavy and it's after lunch
- It assumes trainees (and volunteers) having completed or being familiar with the  Tic-Tac-Toe exercise. Is this a safe assumption.
- It assumes they have knowledge to work with localStorage.

### DIFF
--- a/common-content/en/workshops/breaking-down-a-project/index.md
+++ b/common-content/en/workshops/breaking-down-a-project/index.md
@@ -1,0 +1,126 @@
++++
+title = "Breaking down a project to work in parallel"
+time = 50
+objectives = [
+  "Explain how agreeing a shared data shape and interfaces lets a team work in parallel.",
+  "Break a small requirement into tasks a team can work on in parallel.",
+  "Identify dependencies between tasks and use agreed interfaces to unblock them.",
+]
+[build]
+  render = "never"
+  list = "local"
+  publishResources = false
++++
+
+When a team builds a project together, we break the big job into smaller tasks. But breaking work into small tasks is only half of it. We also have to agree how those tasks fit together, so people can work on them at the same time without blocking each other.
+
+Today we will look at how this worked on a project you have already done, learn the steps to do it on purpose, and then practise on a new one.
+
+## Part 1: How Tic-Tac-Toe let you work in parallel
+
+In the terminal Tic-Tac-Toe project, three people worked at the same time, each on a different file:
+
+- `board-printer.js`
+- `move-maker.js`
+- `status-checker.js`
+
+and `play.js` tied everything together.
+
+Why were you able to work on these at the same time, without waiting for each other? Because two things were agreed **before** anyone started writing code.
+
+1. The shape of the data. The board is always a 3 by 3 array of arrays, holding `'X'`, `'O'` or `'_'`:
+
+```js
+const board = [
+  ["X", "_", "_"],
+  ["_", "X", "_"],
+  ["O", "O", "X"],
+];
+```
+
+Every function in every file expects the board to look exactly like this.
+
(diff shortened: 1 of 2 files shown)


common-content/en/workshops/breaking-down-a-project/index.md
```diff
@@ -0,0 +1,125 @@
...
+const board = [
+  ["X", "_", "_"],
+  ["_", "X", "_"],
+  ["O", "O", "X"],
+];
+```
+
+Every function in every file expects the board to look exactly like this.
+
+**2. The {{<tooltip title="function signatures">}}A function's signature is its name, the parameters it takes, and what it returns. If two people agree a signature, they can each rely on it without seeing the other's code.{{</tooltip>}}.** What each function is called, what it takes, and what it returns:
```
illicitonion: It looks like bolding currently carries through into the tooltip itself, which isn't great. We may want to fix this at some point, but for now let's just not bold this text :)

```suggestion
2. The {{<tooltip title="function signatures">}}A function's signature is its name, the parameters it takes, and what it returns. If two people agree a signature, they can each rely on it without seeing the other's code.{{</tooltip>}}. What each function is called, what it takes, and what it returns:
```

common-content/en/workshops/breaking-down-a-project/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Part 3: Break one down yourselves
+
+Now practise the four steps on a new requirement.
+
+> **Requirement.** Your group is building a small app that keeps a to-do list in the browser. It must:
+>
+> - let a user add a to-do,
+> - show all the to-dos,
+> - and remember the to-dos after the page is reloaded (using `localStorage`).
```
illicitonion: We don't currently teach `localStorage` (we probably should at some point!), so let's maybe reframe this as just having to agree on the format of a global variable, rather than persisting?

illicitonion (commented):
Content-wise, this looks great, I left a couple of small comments.

Location-wise... In theory they start the tic-tac-toe this week, and finish it the next week. They also won't have interacted with the DOM yet, so may struggle with the to do list break-down slightly. I'd suggest moving this to data-flows sprint 2 (where they've just learnt about rendering based on state) and just completely replacing the "Teamwork Project Sprint 2" workshop (which feel quite low value) with this. WDYT?

abdishakoor-dev:
Thanks. Done:
- Moved to Data-Flows Sprint 2, replacing Teamwork Project Sprint 2 (Study Group → 180 to keep the day the same length).
- Un-bolded that line.
- Reframed the practice to a global variable instead of `localStorage`.

Also inlined the `stub`/`hardcode` definitions (they were tooltip-overlapping).

illicitonion (approved):
Looks great, thanks!

Just one more thing - we typically try to always have lunch start at 1pm - how about we extend the demo-qa session by 15 minutes, and add a 15 minute morning break between workshops, so that lunch gets back to 13:00?

abdishakoor-dev:
Good catch! Lunch is now 13:00. Bumped Demo Q&A to 75 and added a 15 min break; also put Study Group back to 150 so the day ends on time.

PR #1943, opened 2026-07-13
Align Functions and Declaring functions LOs with their content

### LINKED ISSUE #1694: Learning Objectives for function prep in ITP
Module: Structuring and Testing Data
Week: 1 (programming fundamentals)

The learning objectives (LOs) feel misleading for two sections. 

The section labelled "[Functions](https://programming.codeyourfuture.io/structuring-data/sprints/1/prep/#functions)" has LOs as stated below:

> Define a function in programming

Can be interpreted as the student learning to define their own function in programming, not a definition of what a function is.

> Evaluate a function expression

Feels largely vague. Does this mean evaluating as in running a function expression? Is it only expected that the student should be able to run existing built-in methods in this stage of the prep material? 

> Call a function with an input in REPL

Again is this only an expectation for built-in functions? Would be great to be clear here.

And for the section called ["Declaring Functions"](https://programming.codeyourfuture.io/structuring-data/sprints/1/prep/#defining-functions) I believe the LO could be more targeted and include one along the lines of `Declare your own function and call it`

It's good to set clear expectations to students on what they're expected to take away from these sections. 

It feels as if the material in "Functions" only cares about built-in functions and not one defined by the student, they aren't expected to understand function syntax yet. But the LOs and name of this section doesn't indicate that clearly enough and could lead students to jump into rabbit holes trying to define their own function without being given the right prep material for it yet.

abdishakoor-dev:
Fixes #1694

Revised the learning objectives for the Functions and Declaring functions blocks so they better reflect the content.

Also reworded the Functions intro so it stands alone rather than carrying on from Evaluating expressions, and noted that `Math.round` is built in.

Preview:
- https://deploy-preview-1943--cyf-curriculum.netlify.app/itp/structuring-data/sprints/1/prep/#functions
- https://deploy-preview-1943--cyf-curriculum.netlify.app/itp/structuring-data/sprints/1/prep/#declaring-functions

### DIFF
--- a/common-content/en/module/js1/declaring-functions/index.md
+++ b/common-content/en/module/js1/declaring-functions/index.md
@@ -4,6 +4,7 @@ title = 'Declaring functions'
 time = 20
 [objectives]
     1='Label and describe the syntactic features of a function declaration'
+    2='Declare your own function using a function declaration'
 [build]
   render = 'never'
   list = 'local'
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Thanks!

PR #1947, opened 2026-07-15
Add launch projects

illicitonion:
@kfklein15 I added tech stack requirements, both to make them easier to maintain, and to try to push towards different tech stacks. For the career mentor matching, I told them to use TypeScript to make it easier to potentially integrate into the Portal in the future if we want to.


kfklein15 (approved):
LGTM

Issue #1946, opened 2026-07-15
Cover error handling: exceptions, try/catch, and when to throw


abdishakoor-dev:
**Which module(s) and week(s) does this change affect?**
Module(s): ITP, earlier than Data Flows (e.g. Structuring Data or Data Groups), plus Data Flows Sprint 3 for the async follow-up
Week(s): To be decided

**What is the work that needs to be done?**

We want to properly cover error handling:

- What exceptions are and what happens when one goes unhandled
- What try/catch is and how to use it
- When you should throw vs handle the error another way (return a value, log, recover, re-throw)
- How to handle errors well: catching at the right level, not swallowing errors, giving users/callers useful information

There is an existing block, `common-content/en/module/js3/catch`, which covers try/catch around async fetch calls. It was deliberately removed from Data Flows Sprint 3 prep in 4f8508cd (#1111), with this rationale in the commit message:

> 1. We haven't actually taught `try-catch` anywhere, so we can't just rely on transferring "it's the same as you already know, just async".
> 2. There are already _way_ too many novel concepts in Data Flows Sprint 3, and it's already very hard to follow from scratch.
>
> If we add `try-catch` earlier in the course (which I think we should), we should re-add this.

So the work is likely two parts:

1. Introduce error handling fundamentals earlier in the course, in a synchronous context
2. Re-add the `js3/catch` block (or an updated version) to Data Flows Sprint 3, building on that foundation

**Why is this work important to do?**

Trainees currently learn fetch, promises, and async/await without ever being taught what to do when things fail. The removal in #1111 was explicitly conditional on teaching this earlier, which hasn't happened yet.

**Who might need to know about this change?**

@illicitonion

jyothikiran09:
I support introducing error handling earlier in the curriculum.

Learning `fetch`, Promises, and `async/await` without first understanding exceptions and `try...catch` makes it difficult for learners to know how to respond when something goes wrong. Teaching the fundamentals first would make the later asynchronous concepts much easier to understand.

I suggest covering:

* What exceptions are and what happens when they are unhandled.
* How to use `try...catch` with simple synchronous examples.
* When to `throw` an error versus returning a value, logging the error, recovering from it, or re-throwing it.
* Good error-handling practices, such as catching errors at the appropriate level, avoiding swallowed errors, and providing meaningful error messages.

[...]

PR #1949, opened 2026-07-22
Add error handling fundamentals to Data Groups Sprint 1

abdishakoor-dev:
First attempt for #1946

Adds three js2 blocks (Throwing errors, try/catch, Throw or return?) that teach exceptions and error handling using the existing calculateMedian exercise, and wires them into Data Groups Sprint 1
prep, after "Implementing all the cases".

Trainees currently learn fetch, promises, and async/await without ever being taught how to throw or catch an error. This covers the synchronous foundation.

Two follow ups needed once this lands:
- Update [Module-Data-Groups: Sprint-1/fix](https://github.com/CodeYourFuture/Module-Data-Groups/blob/main/Sprint-1/fix) currently expects calculateMedian to return null for invalid input, however, we are now teaching in this prep throwing errors
- Re-add the async try/catch block to Data Flows Sprint 3 (removed
  in #1111)

Preview: https://deploy-preview-1949--cyf-curriculum.netlify.app/itp/data-groups/sprints/1/prep/#throwing-errors


common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+When a function cannot do its job, it shouldn't guess. It should fail immediately and loudly, at the point where the problem is. We call this {{<tooltip title="failing fast">}}Failing fast means stopping as soon as we know something is wrong, instead of carrying on with bad data and failing later somewhere confusing.{{</tooltip>}}.
+
+### Using `throw`
+
+In Structuring Data, you [interpreted error traces](/itp/structuring-data/sprints/2/prep/#interpreting-this-error) when JavaScript threw a `SyntaxError` at you. Now we can {{<tooltip title="throw">}}Throwing an error stops normal execution immediately. The error travels up through the calling functions until something handles it. If nothing handles it, the program crashes and prints the error trace.{{</tooltip>}} errors from our own code, using the `throw` keyword:
```
illicitonion: This link works, but could plausibly break quite easily (e.g. if that block was re-titled) and we would be unlikely to notice (the link would just start going to the first prep block). Accordingly, we generally try to avoid linking directly to specific prep blocks.

I don't love that. Maybe we could at some point add a "check if the anchors linked to actually exist" check so we'll notice if they break and can fix them.

I'm not sure I'm suggesting a specific action here, but just an FYI...

illicitonion:
```suggestion
In Structuring Data, you [interpreted error traces](/itp/structuring-data/sprints/2/prep/#interpreting-this-error) when JavaScript threw a `SyntaxError` at you. We can also {{<tooltip title="throw">}}Throwing an error stops normal execution immediately. The error travels up through the calling functions until something handles it. If nothing handles it, the program crashes and prints the error trace.{{</tooltip>}} errors from our own code, using the `throw` keyword:
```

common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+```js
+function calculateMedian(list) {
+  if (Array.isArray(list) === false) {
```
illicitonion:
```suggestion
  if (!Array.isArray(list)) {
```

common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+    throw new Error("calculateMedian requires an array of numbers");
+  }
+  if (list.length === 0) {
+    throw new Error("calculateMedian requires a non-empty array");
```
illicitonion, line 70: We don't check they're all numbers - do we want to?

common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+We can assert that a function throws using the [`toThrow`](https://jestjs.io/docs/expect#tothrowerror) matcher:
+
+```js
+test("throws an error when given an empty list", () => {
+  expect(() => calculateMedian([])).toThrow();
+});
+```
+
+Note that we wrap the call in a function. If we called `calculateMedian([])` directly, the error would be thrown before `expect` could check anything.
```
illicitonion:
```suggestion
Note that we wrap the call in a function. When we call a function with arguments, those arguments get evaluated before the function is called. If we called `expect(calculateMedian([])).toThrow()` directly, the error would be thrown before `expect` was called, and so before `expect` could check anything.
```

common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+Now the error points at the exact place the problem was detected, with a message saying what the problem is. That's much more useful to whoever calls our function than a mysterious `NaN`.
+
+### Testing that a function throws
+
+We can assert that a function throws using the [`toThrow`](https://jestjs.io/docs/expect#tothrowerror) matcher:
+
+```js
+test("throws an error when given an empty list", () => {
+  expect(() => calculateMedian([])).toThrow();
```
illicitonion: Shall we add an error message (fragment) to the `toThrow()` call, to show people a better example of usage?

common-content/en/module/js2/throwing-errors/index.md
```diff
@@ -0,0 +1,84 @@
...
+  }
+  const middleIndex = Math.floor(list.length / 2);
+  if (list.length % 2 === 0) {
+    return (list[middleIndex - 1] + list[middleIndex]) / 2;
+  }
+  return list[middleIndex];
+}
+```
+
+[`Array.isArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) checks whether a value is an array. If the input isn't an array, or is an empty one, we `throw` a new [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) with a message explaining what went wrong.
```
illicitonion: I think it's probably worth an aside talking about what a good error message looks like?

e.g. a common piece of feedback I give in code reviews is "Include an indication of what the bad data was"

common-content/en/module/js2/throw-or-return/index.md
```diff
@@ -0,0 +1,63 @@
...
+  list = 'local'
+  publishResources = false
+
++++
+
+We now have two ways for a function to respond when it meets a problem: return a value, or throw an error. How do we choose?
+
+### Expected situations: use a return value
+
+If a situation is a normal part of using the function, handle it with ordinary code. Searching an array for a value that isn't there is not an error; that's why [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) returns `-1` rather than throwing. The caller expects both outcomes and can check with an `if`.
```
illicitonion, line 19: Let's add some guidance on what value to choose, maybe with some examples? e.g. that `-1` is a good thing to return from `indexOf` because it probably fits in with how it's used (whereas returning `null` would be more annoying), that string methods may want to return `""` or `null` depending on context, etc?

common-content/en/module/js2/throw-or-return/index.md
```diff
@@ -0,0 +1,63 @@
...
+>}}
+
+{{<multiple-choice
+  question="A user types 'eleven' into a form field asking how many tickets they want. What should the program do?"
+  answers="Check the input and show the user a helpful message | Throw an error | Nothing, the input is fine"
+  feedback="Right. Users typing unexpected things is normal human behaviour, not a bug, so handle it with ordinary conditional logic. | A person mistyping is an expected situation, not a broken expectation in our code. Throwing turns normal use into a crash. | 'eleven' is not a number, so the program cannot book eleven tickets without dealing with it somehow."
+  correct="0"
+>}}
+
+{{<multiple-choice
```
illicitonion, line 53: I'd maybe answer this with "it depends"? Sometimes "continue despite bad data" is a useful answer for a user and sometimes "error if _anything_ is wrong" is a useful answer...

common-content/en/module/js2/throw-or-return/index.md
```diff
@@ -0,0 +1,63 @@
...
+  answers="Catch the error, report the bad file, and continue | Let the whole program crash | Silently skip the file"
+  feedback="Right. The loop can do something useful with the error: skip the bad file, tell the user about it, and still summarise the other 99. | Crashing throws away 99 files of good work when the loop could recover and report the problem instead. | This swallows the error: the summary is quietly missing a file and nobody knows to fix it."
+  correct="0"
+>}}
+
+{{<multiple-choice
+  question="divide(a, b) is called with b equal to 0. What should divide do?"
+  answers="Throw an error | Return 0 | Return Infinity because that is what JavaScript does"
+  feedback="Right. divide cannot do what its name promises with a zero divisor, and a zero here usually means a bug in the calling code. Some teams make a different, documented choice, but it must be a deliberate one. | Zero is a made-up answer that hides the caller's bug and will be used as if it were real. | JavaScript's own operator does evaluate to Infinity, but for our function that quietly passes a strange value along instead of surfacing the problem where it happened."
```
illicitonion:
```suggestion
  feedback="Right. divide cannot do what its name promises with a zero divisor, and a zero here usually means a bug in the calling code. Some teams make a different, documented choice, but it must be a deliberate one. | Zero is a made-up answer that hides the caller's bug and will be used as if it were real. | JavaScript's own operator does evaluate to Infinity, but for our function that would quietly pass a strange value along instead of surfacing the problem where it happened."
```

illicitonion (commented):
LGTM! Left a few minor comments.

Exercise-wise, as well as updating the existing ones, do we want to add a few new exercises to get people throwing and catching more?

Also, do we want to add a callback to this in https://curriculum.codeyourfuture.io/sdc/tools/sprints/5/prep/ to reflect on static checking being better than dynamic checking?

abdishakoor-dev:
I've addressed all the inline comments in the last few commits. 

Preview: https://deploy-preview-1949--cyf-curriculum.netlify.app/itp/data-groups/sprints/1/prep/#throwing-errors

On the two follow-up questions, I've filed separate issues to track them rather than block this PR:

- **New/updated exercises**: [Module-Data-Groups#1443](https://github.com/CodeYourFuture/Module-Data-Groups/issues/1443)
- **Static vs dynamic checking callback**: [curriculum#1987](https://github.com/CodeYourFuture/curriculum/issues/1987)

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: Removed the link for now to be on safe side and to be ever green

abdishakoor-dev: Done

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: Agreed. Done.

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: Done

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: Done.

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: Agreed. Done.

common-content/en/module/js2/throwing-errors/index.md
abdishakoor-dev: I have expanded into it based on your suggestion

common-content/en/module/js2/throw-or-return/index.md
abdishakoor-dev: Done

common-content/en/module/js2/throw-or-return/index.md
abdishakoor-dev: I have to choose answer as the valid answer in the quiz question but I have expanded the answer to give a more rounded answer

common-content/en/module/js2/throw-or-return/index.md
abdishakoor-dev: Done

illicitonion (approved):
LGTM - thanks!

PR #1954, opened 2026-07-22
ambigious instructions

mjpeet:
/sdc/tools/sprints/3/day-plan/ - strace

clear up some ambiguity for trainees (originally this rendered on one line, trainees were copying out additional text thinking that's part of the command)


illicitonion (approved):
Thanks!

PR #1966, opened 2026-07-27
PR needing review page should use create date not update date

LonMcGregor:
Only using the most recent update date gives a false impression of our review status.

This is a problem if a trainee has an old review but they toggled the "needs review" label in an attempt to get our attention, as I saw happen recently.

We should always use the oldest date here.

Thoughts?

### DIFF
--- a/org-cyf-theme/assets/custom-scripts/reviews/index.mjs
+++ b/org-cyf-theme/assets/custom-scripts/reviews/index.mjs
@@ -99,14 +99,14 @@ function render() {
       continue;
     }
 
-    awaitingReviewByAge[pr.module][pr.updatedAge]++;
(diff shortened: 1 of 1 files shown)


illicitonion (commented):
I've been modelling this as "time spent waiting" rather than "time since it was created"...

Would a best of both worlds be to be checking label _transitions_? i.e. "When was Needs Review last added when before it either didn't have any relevant label, or had the label Reviewed"?

I don't mind making this change as-is if it's simpler and useful, but I suspect "I opened a PR a long time ago but only just acted on feedback and re-requested review" is a more common occurrence than "I toggled the Needs Review label because I got frustrated"

LonMcGregor:
That makes sense, after reflecting for a bit this change is probably not needed. I'll close this

PR #1969, opened 2026-07-28
Change mermaid version from @latest to @11.16.0

cjyuan:
## Changelist
- Changed Mermaid CDN import from `@latest` to the a pinned version 11.16.0 to avoid potential 404 errors (which causes diagrams not showing).


abdishakoor-dev (approved):
Approving. Diagrams render correctly on the deploy preview:
https://deploy-preview-1969--cyf-curriculum.netlify.app/itp/onboarding/sprints/1/prep/#working-locally

The root cause appears to be that jsDelivr does not serve `dist/chunks/**` under the `@latest` alias, so Mermaid's lazy imports 404 on production. Pinning fixes it.

PR #1972, opened 2026-07-31
Render sdc issues from Portal not GitHub

illicitonion:
We still render the same information, including GitHub links.

The only significant change in output is that we now start rendering
times properly rather than falling back to a label, because we have this
structured information now, and rendering the task code if known.

The Portal doesn't have arbitrary labels, so we manufacture a single
category label ourselves. We don't currently render the labels anywhere
other than when we don't know the time of an issue, so this feels fine.

In the future we will:
1. Swap all courses over to use this style.
2. Probably start removing some of the GitHub-dependent things (like
   links to issues, and clone buttons)
3. Clean up some of the indirection when we're back down to one way of
   doing things.


cifarquhar (approved):
Awesome, thanks!

PR #1973, opened 2026-07-31
Display time as hours if big enough

illicitonion:
We have some multi-hour courseworks, and "16 hours" reads a lot better than "960 minutes".


cifarquhar (approved):
LGTM

PR #1994, opened 2026-08-26
Update Checkpoint docs

illicitonion:
Remove granular day-plans. We don't actually follow them. Instead, give "whole day" day plan items, with descriptions of what is expected during the day.

Also:
* Remove breaking down projects workshop - it's now covered in ITP.
* Spec that projects should be written in JavaScript.
* Clarify that you don't need a definite interview outcome.
* Guide towards scheduling demos in advance of the day.

### DIFF
--- a/common-content/en/module/checkpoint/solo-project/index.md
+++ b/common-content/en/module/checkpoint/solo-project/index.md
@@ -1,6 +1,5 @@
 +++
 title="Solo Project"
-time=8640
 [build]
   render = 'never'
   list = 'local'
--- a/common-content/en/module/checkpoint/solo-project/index.md
+++ b/common-content/en/module/checkpoint/solo-project/index.md
@@ -1,10 +1,6 @@
 +++
 title="Solo Project"
-time=45
-tasks = [
-    "Pick a project to work on on your own.",
-    "Start working on a solo project.",
-]
+time=8640
 [build]
   render = 'never'
   list = 'local'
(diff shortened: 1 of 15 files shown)


common-content/en/module/checkpoint/solo-project/index.md
```diff
@@ -1,10 +1,6 @@
 +++
 title="Solo Project"
-time=45
-tasks = [
-    "Pick a project to work on on your own.",
-    "Start working on a solo project.",
-]
+time=8640
```
Liam310: This equates to 144 hours, which I assume you've done because that's how long it is until the submission deadline, but I think maybe it looks a little more like that's how many hours of work the project is expected to take. What do you think?

Image

Liam310: (I recognise this is the same as it is for the group project, so I'm really asking whether we ought to change it for those, too)

Liam310:
Looks great! I appreciate the streamlining 🙂 

I've left a single comment but am happy to approve the PR and then let you decide whether it's worth actioning. Sadly, I currently _can't_ approve the PR because I don't have permissions still 😅

common-content/en/module/checkpoint/solo-project/index.md
illicitonion: Hidden :)

Liam310 (approved):

PR #1996, opened 2026-08-28
Update Launch demo guidance

illicitonion:


kfklein15 (approved):
Good changes to clarify to trainees what is expected on the demo.

Can you please let Azin know demos will be 15 min + 5 min questions? This impacts her planning of the agenda. 

@cifarquhar can you share this with the trainees too, please?

PR #1997, opened 2026-08-29
docs(employability): add an optional final CV self-check

MuslumYilmaz:
## Summary
Add one optional self-check after the CV formatting guidance, linking to a free frontend-focused linter.

## Why
This gives learners a concrete way to check ATS-style structure, parsing, readability, frontend keywords, and evidence after applying the page’s advice. The tool uses deterministic checks, does not rewrite achievements, and states that uploaded files are not stored.

## Disclosure
I’m affiliated with FrontendAtlas, which maintains the linked free tool.

## Validation
The Markdown placement, rendered preview, and live destination were checked. This is a documentation-only change; project CI can validate formatting and links.


illicitonion:
Thanks for the suggestion! We don't specifically train frontend engineers, so linking to a frontend-specific tool here probably doesn't make sense - if there's a non-frontend-specific version of this tool I'd be happy to try it out and see if we want to recommend it.

PR #2001, opened 2026-09-03
Add Prep quiz block to ITP day plans

abdishakoor-dev:
Adds a 15 minute Prep quiz block after morning orientation in every ITP day plan (onboarding and all 12 sprints). The facilitator runs the sprint's QuizIt quiz as a recap of the prep, and notes anything the room got wrong for Study Group.

Morning orientation drops from 15 to 5 minutes and energisers lose 5, so days' length stays the same.

Preview:
- https://deploy-preview-2001--cyf-curriculum.netlify.app/itp/onboarding/day-plan/#prep-quiz
- https://deploy-preview-2001--cyf-curriculum.netlify.app/itp/data-flows/sprints/1/day-plan/#prep-quiz
- https://deploy-preview-2001--cyf-curriculum.netlify.app/itp/testing/sprints/1/day-plan/#prep-quiz

### DIFF
--- a/common-content/en/blocks/prep-quiz/index.md
+++ b/common-content/en/blocks/prep-quiz/index.md
@@ -1,0 +1,31 @@
++++
+title = 'Prep quiz'
+description = 'Recap the prep as a quiz and find out what the room is stuck on'
+time = 15
+hide_from_overview = true
+[build]
+  render = 'never'
+  list = 'local'
+  publishResources = false
+
++++
+
+{{<note title="Prep quiz (15 minutes)" type="activity">}}
+
+This is the recap for the sprint, run as a quiz. Everyone has done the prep during the week. A quiz brings it back to the front of everyone's mind, gets the whole room involved rather than just listening, and shows straight away which parts people are still unsure of.
+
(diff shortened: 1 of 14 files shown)


illicitonion (commented):
Generally LGTM but a few questions:
1. Currently I only see a few quizzes - what's the plan to end up with one per sprint?
2. How is quizit deployed/developed/maintained? If I wanted to edit a quiz, how would I?
3. Accessibility-wise on quizit, the right/wrong answer appears to currently only be indicated by a green/red highlight? For colour-blind users that may be an issue - can we also add correct/incorrect text?
4. I've noticed that often "the longest answer" is the correct answer - in the future when adding questions let's try to keep answers similar lengths and levels of detail to avoid being able to guess based on this.
5. When I got to the end of a quiz my score was "8700 pts" - I'm not sure what points are? It feels like "17/20" would be more useful feedback?

And for _this_ sprint:
 * "What must be true by the end of week 6 for you to enrol as a trainee?" - "You have attended at least three classes" feels like also at least somewhat correct?
 * "Help with childcare, travel, broadband, food in class and a laptop loan, for the first 5 weeks until you enrol as a trainee" let's update to 6, and maybe rephrase to "and after you enrol as a trainee"?

PR #2014, opened 2026-09-14
ITP: Emphasise doing prep work on separate branch

cjyuan:
## Changelist
- Add an instruction to instruct trainee to carry out prep work in a separate branch (to prevent them from polluting their `main`)


cifarquhar (approved):
Great spot, thank you!

